PR #10, opened 2019-01-07
Add documentation based on sphinx

xadupre:
And also:

* remove unnecessary functions
* rename convert_sklearn into convert


.gitignore
```diff
@@ -30,3 +30,7 @@ coverage.xml
 .coverage
 __dump_data/
 junit/
+docs/auto_examples
+docs/examples/graph*.*
+docs/examples/rf_iris.onnx
```
wenbingl: It's kind of not general

PR #12, opened 2019-01-08
Remove the obsolete travis ci file.

wenbingl:


.azure-pipelines/win32-conda-CI.yml
```diff
@@ -36,12 +36,10 @@ jobs:
       python -m pip install --upgrade pip
       pip install -r requirements.txt
       pip install -r requirements-dev.txt
-      test '$(python.version)' != '2.7' && pip install onnxruntime
```
prabhat00155: Why are you removing this from here? This is present in linux-conda-CI.yml.

.azure-pipelines/win32-conda-CI.yml
```diff
@@ -36,12 +36,10 @@ jobs:
...
       pip install pytest
     displayName: 'Install dependencies'
 
   - script: |
-      test '$(python.version)' != '2.7' && python -c "import onnxruntime"
```
prabhat00155: Same comment as above.

.azure-pipelines/win32-conda-CI.yml
wenbingl: No way to python 2.7 in Windows since there is no available onnx package for that.

PR #14, opened 2019-01-09
Make tests folder part of the package.

xadupre:
Tests should be part of the package to allow users to test a converter against a specific version of the runtime. Other packages implementing converters could leverage the tests implemented here.


wenbingl:
No, don't do that. It's very unusual to put test stuff into the package, and consider the goal of this splitting work.
If the user really want to use something from the test folder, he should enlist git repro and do his work.

xadupre:
Scikit-learn does that, nimbusml does that and i would like to have the same testing code in skl2onnx and onnxmltools. I don't mind creating another package but i think we need common tools to develop converters.

Think of this scenario: you need to deploy somewhere and it works. You install skl2onnx, onnxruntime and you do skl2onnx.check_converters(). Then you know everything perfectly works. Git clone is ok but you always get the dev version.

wenbingl:
Your scenario could be applied into every pip package, but when you search your pip package folder you will see most of them actually did not pack the test code. it's not the user responsibility to be our project QA. And even you probably not run any test code from your installed pip packages. If you still think the test part is very useful for the end user, you can publish a test package to see how many users will really install it. 

One related valid scenario is to check the compatibility between the converter and the runtime. The recent discussion concluded there would be a common project to put all of these test utils, optimizer and etc. You can create a sub folder with setup.py in onnxmltools to check in your test utils.

xadupre:
The first thing users do to see if onnxruntime is useful is to compare execution time between onnxruntime and scikit-learn (or any other library). We should propose such functionality soon. The code I moved to the main subfolder is related to that functionality. It is used in skl2onnx and onnxmltools. I'd prefer to create a new repo with a proper build and documentation rather than create a temporary subfolder in onnxmltools. I moved this code in the main subfolder because sklearn is usually a dependency so that's seemed convenient to have some code to benchmark models here.

xadupre:
This will be designed in a different way.

PR #22, opened 2019-01-15
Add converters for KMeans, MiniBatchKMeans

xadupre:


skl2onnx/operator_converters/KMeans.py
```diff
@@ -0,0 +1,122 @@
...
+                        name=scope.get_unique_operator_name('ReduceSumSquare'))
+    
+    # Compute -2XC'
+    zero_name = scope.get_unique_variable_name('zero')
+    zeros = np.zeros((N, ))
+    container.add_initializer(zero_name, onnx_proto.TensorProto.FLOAT,
+                              list(zeros.shape), zeros)   
+    
+    if container.target_opset < 5:
+        attrs['broadcast'] = 1
```
wenbingl: where is attrs used?

skl2onnx/operator_converters/KMeans.py
```diff
@@ -0,0 +1,122 @@
...
+        op_version = 1
```
wenbingl: Given this is first time to support this op, probably we don't care about opset less than 7.

xadupre: I moved all operators which dépends on version in _apply_operator. kmeans and gemm operator should be two independant problem now.

PR #24, opened 2019-01-16
Make sure the code is using apply_function when available

xadupre:
Functions in _apply_operation. Most of these functions have different signature depending on target_version.

### DIFF
--- a/skl2onnx/common/_apply_operation.py
+++ b/skl2onnx/common/_apply_operation.py
@@ -6,8 +6,8 @@
 # This file contains some high-level APIs for applying operations on variables specified by names. We should try our
 # best to use those functions because they can produce ONNX operators according to the ONNX version specified in the
 # `container` argument. Notice that those function behaviors are defined in a way very similar to ONNX-1.2.
-
 from ..proto import onnx_proto
+
 
 def _create_name_or_use_existing_one(scope, op_type, name):
     if name is None:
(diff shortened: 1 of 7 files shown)


skl2onnx/common/_container.py
```diff
@@ -3,10 +3,41 @@
...
 # --------------------------------------------------------------------------
-
+import inspect
+import re
+import traceback
 import six
+import sys
 from ..proto import helper
 from .interface import ModelContainer
+from ._apply_operation import __dict__ as dict_apply_operation
```
wenbingl: a little hacky, why not just import it and dir all items?

tests/test_applied_functions.py
```diff
@@ -0,0 +1,19 @@
+"""
+Tests scikit-learn's binarizer converter.
```
wenbingl: wrong comments

skl2onnx/common/_apply_operation.py
```diff
@@ -221,21 +221,23 @@ def apply_reciprocal(scope, input_name, output_name, container, operator_name=No
...
-    else:
+    elif desired_shape:
         # The shape attribute of Reshape becomes a tensor input, so we create one tensor to store that attribute.
         desired_shape_name = scope.get_unique_variable_name('shape_tensor')
         container.add_initializer(desired_shape_name, onnx_proto.TensorProto.INT64, [len(desired_shape)], desired_shape)
 
         # Create ONNX Reshape operator
         container.add_node('Reshape', [input_name, desired_shape_name], output_name, op_version=5, name=name)
+    else:
+        container.add_node('Reshape', input_name, output_name, name=name)
```
wenbingl: no opset number here?

skl2onnx/common/_container.py
xadupre: pylint raises a warning for from … import * so I usually avoid it.

skl2onnx/common/_apply_operation.py
```diff
@@ -221,21 +221,23 @@ def apply_reciprocal(scope, input_name, output_name, container, operator_name=No
 
 
 def apply_reshape(scope, input_name, output_name, container, operator_name=None, desired_shape=None):
-    if len(list(i for i in desired_shape if i < 0)) > 1:
-        raise ValueError('There can only be one -1 in the targeted shape of a Reshape but got %s' % desired_shape)
 
     name = _create_name_or_use_existing_one(scope, 'Reshape', operator_name)
 
     if container.target_opset < 7:
```
vinitra: Should this be container.target_opset < 5 since reshape was most recently updated in opset 5?

xadupre: It is taken from onnxmltools. I assume it was for a good reason. The build set up on travis was trying other version of ONNX. We should do that again.

PR #56, opened 2019-02-14
Fixed Naives bayes converter/unit tests

prabhat00155:


tests/test_SklearnNaiveBayesConverter.py
```diff
@@ -1,63 +1,78 @@
...
-                            allow_failure="StrictVersion(onnxruntime.__version__) <= StrictVersion('0.1.3')")
+        dump_data_and_model(X, model, model_onnx,
+                        basename="SklearnMclMultinomialNB",
+                        allow_failure="StrictVersion(onnxruntime.__version__)"
+                                       "<= StrictVersion('0.2.1')")
 
-    @unittest.skip(reason="Type 'tensor(float)' of input parameter (input) of operator (ConstantOfShape) "
-                          "in node (ConstantOfShape) is invalid.")
+    @unittest.skipIf(StrictVersion(onnx.__version__) <= StrictVersion('1.3'),
+                     'Needs greater opset 9')
```
wenbingl: reason statement doesn't align with '(StrictVersion(onnx.__version__) <= StrictVersion('1.3'),'

prabhat00155: How does it not align? onnx version 1.3 corrsponds to opset 8 (https://github.com/onnx/onnx/blob/master/docs/Versioning.md#released-versions). I am checking here if onnx.__version__ <= 1.3 and if it is, then skip the test with the message 'needs greater opset 9 (that corresponds to onnx.__version__ 1.4.1)'.

wenbingl: Sorry, I always was confused by the statement is actually opposite to the conditional statement.

PR #62, opened 2019-02-19
additional option for ONNX operators, fix bugs in pipeline conversionn changes default domain (no empty value anymore)

xadupre:


tests/test_SklearnPipelineWithinPipeline.py
```diff
@@ -62,7 +63,8 @@ def test_pipeline_pca_pipeline_none_lin(self):
                                      [('input', FloatTensorType(data.shape))])
         self.assertTrue(model_onnx is not None)
         dump_data_and_model(data, model, model_onnx,
-                            basename="SklearnPipelinePcaPipelineMinMaxLogReg")
+                            basename="SklearnPipelinePcaPipelineMinMaxLogReg",
+                            allow_failure="StrictVersion(onnxruntime.__version__) <= StrictVersion('0.2.1')")
```
wenbingl: instead of add this conditional check everywhere, should we group these tests into sub-folder? it looks the most sklearn converter has to be opset 9 now.

xadupre: Let's add a new issue then?

docs/examples/plot_tfidfvectorizer.py
```diff
@@ -0,0 +1,190 @@
...
+            # ('body_stats', Pipeline([
+            #     ('stats', TextStats()),  # returns a list of dicts
+            #     ('vect', DictVectorizer()),  # list of dicts -> feature matrix
+            # ]), 1),
+        ],
+
+        transformer_weights={
+            'subject': 0.8,
+            'body_bow': 0.5,
+            # 'body_stats': 1.0,
```
prabhat00155: Why do you have this commented out here? If you don't need it, may be remove the line.

skl2onnx/operator_converters/Flatten.py
```diff
@@ -0,0 +1,20 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+from ..proto import onnx_proto
+from ..common._apply_operation import apply_cast, apply_concat
```
prabhat00155: Why do you import these two when it is not being used?

skl2onnx/operator_converters/Flatten.py
```diff
@@ -0,0 +1,20 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+from ..proto import onnx_proto
```
prabhat00155: Where do you use onnx_proto?

docs/examples/plot_tfidfvectorizer.py
xadupre: I hesitated to keep it. It comes from the original examples. I think I wrote the example was more simple than the one from the documentation because it is not yet easy to convert such custom transformer but I let it to let the user know that's missing from the original example.

skl2onnx/operator_converters/Flatten.py
xadupre: I did not see. I usually implement a unit test checking pep8 on my open source projects which I run before checking any change. So I forget to check that now. My mistake.

PR #64, opened 2019-02-22
Modified scikit converter code to align with pep-8

prabhat00155:


skl2onnx/shape_calculators/SVM.py
```diff
@@ -29,31 +29,16 @@ def calculate_sklearn_svm_output_shapes(operator):
...
         elif all(isinstance(i, (numbers.Real, bool, np.bool_)) for i in op.classes_):
             operator.outputs[0].type = Int64TensorType([N])
-            if len(operator.outputs) == 2:
-                if operator.target_opset < 7:
-                    # Old ONNX ZipMap produces Map type
-                    operator.outputs[1].type = DictionaryType(Int64TensorType([1]), FloatTensorType([1]))
-                else:
-                    # New ONNX ZipMap produces Seq<Map> type
-                    operator.outputs[1].type = \
-                        SequenceType(DictionaryType(Int64TensorType([]), FloatTensorType([])), N)
```
wenbingl: why do you change this output format?

wenbingl:
any tool to automatically reformat the code into 72 chars per line?

skl2onnx/shape_calculators/SVM.py
prabhat00155: We moved Zipmap to the parsing step. I forgot to change SVM classifier's output to return a probability tensor, when I changed output type of other classifiers.

skl2onnx/operator_converters/CalibratedClassifierCV.py
```diff
@@ -46,61 +48,61 @@ def convert_calibrated_classifier_base_estimator(scope, operator, container, mod
     # ARRAYFEATUREEXTRACTOR <- k [1] -> ARRAYFEATUREEXTRACTOR
     #         |                          |
     #         V                          V
-    #  transposed_df_col[M, 1]        transposed_df_col[M, 1]
```
xadupre: I think we should increase the limit on line size. I think the graph below should be part of the docstring and not in comment. I'm ok to break the limit here if it is easier to read.

PR #65, opened 2019-02-25
Aligned shape_calculators code to pep8

prabhat00155:


skl2onnx/shape_calculators/__init__.py
```diff
@@ -19,11 +20,36 @@
...
-from . import SVM
 from . import SVD
+from . import SVM
 from . import TextVectorizer
 from . import TfidfTransformer
 from . import VotingClassifier
-from . import ZipMap 
+from . import ZipMap
+
+__all__ = [
```
xadupre: Do we need that since it is the same list as above?

prabhat00155: Yes, we need to define a list of public objects of a module. For example:
https://github.com/scikit-learn/scikit-learn/blob/7389dbac82d362f296dc2746f10e43ffa1615660/sklearn/svm/__init__.py

wenbingl: Given the 'from xxx import *' is not encouraged in pylint, __all__ seems not necessary any more.
Also, if so, need be consistent on the folder operator_converters

prabhat00155: I have added it to operator_converters as well. Also, it would give a warning in flake8 without this for unused imports.

PR #66, opened 2019-02-26
pep8 changes in skl2onnx and add flake8 check to ci build

prabhat00155:


skl2onnx/_supported_operators.py
```diff
@@ -28,18 +28,18 @@
 from sklearn.multiclass import OneVsRestClassifier
 
 # Tree-based models
+from sklearn.ensemble import ExtraTreesClassifier
```
wenbingl: import as 'alphabetic order' is recommended in PEP8?

skl2onnx/__init__.py
```diff
@@ -15,9 +15,9 @@
 __model_version__ = 0
 
 
-from .convert import convert_sklearn
-from ._supported_operators import update_registered_converter
-from ._parse import update_registered_parser
+from .convert import convert_sklearn # noqa
```
wenbingl: it looks a very normal import, why # noqa?

prabhat00155: This import is not directly getting used in this file, so flake8 throws a warning.

skl2onnx/_supported_operators.py
prabhat00155: It's not mentioned in PEP8, but it improves readability.

prabhat00155: It's not mentioned in PEP8, but it improves readability.

PR #69, opened 2019-03-02
Prroy/adaboost

prabhat00155:


skl2onnx/operator_converters/AdaBoost.py
```diff
@@ -0,0 +1,208 @@
...
+    container.add_initializer(
+        est_weights_sum_name, onnx_proto.TensorProto.FLOAT,
+        [], [model.estimator_weights_.sum()])
+    container.add_initializer(
+        mul_operand_name, onnx_proto.TensorProto.FLOAT,
+        [], [1. / (model.n_classes_ - 1)])
+    container.add_initializer(zero_scaler_name,
+                              onnx_proto.TensorProto.INT32, [], [0])
+
+    container.add_node('Sum', proba_names_list,
```
wenbingl: for some common op like 'Sum', should we create apply_sum function for it?

xadupre: My understanding was operators Added with apply_* have different opset versions.

wenbingl: @xadupre @prabhat00155 , we never know which op will be updated in the future and don't want to update the existing code for a new versioned op on adding a new apply_xxx. So understanding is the frequency of op.

PR #73, opened 2019-03-06
Added AdaBoostRegressor converter and its unit test

prabhat00155:


tests/test_SklearnAdaBoostConverter.py
```diff
@@ -41,14 +42,31 @@ def test_ada_boost_classifier_samme(self):
...
-                                      FloatTensorType(X_train.shape))])
+                                      FloatTensorType(X_test.shape))])
         self.assertIsNotNone(model_onnx)
         dump_data_and_model(X_test.astype('float32'), model, model_onnx,
                             basename="SklearnAdaBoostClassifierSAMME",
                             allow_failure="StrictVersion("
                                           "onnxruntime.__version__)"
                                           "<= StrictVersion('0.2.1')")
 
+    @unittest.skip(reason="CumSum op has not yet been implemented in onnx")
```
wenbingl: Do you know if there is anyone is working on this onnx Op? If not, we should not enable this kind of converter which can't run with any onnx-runtime.

wenbingl:
@prabhat00155, In general, the converted model should always be able to run with onnx-runtime, otherwise, we should throw an exception during the conversion. This is an agreement among teams.

tests/test_SklearnAdaBoostConverter.py
prabhat00155: I created this task: https://aiinfra.visualstudio.com/Lotus/_workitems/edit/2740. It is still in unassigned state.

prabhat00155:
In this case, I need an op to calculate Cumulative Sum (https://aiinfra.visualstudio.com/Lotus/_workitems/edit/2740). What do you suggest I do here, wait for onnx to implement the ops, which would take time. Currently the runtime would throw an error.

PR #74, opened 2019-03-07
opset 9 update for cast op

prabhat00155:


skl2onnx/common/_apply_operation.py
```diff
@@ -138,17 +138,30 @@ def apply_cast(scope, input_name, output_name, container, operator_name=None,
         v.number: k for k, v in d.values_by_name.items()
     }
     if to not in allowed_type_name_and_type_enum_pairs:
-        raise ValueError('Attribute to must be one of %s'
```
xadupre: It is usually "Attribute 'op' must...

prabhat00155: PEP8 doesn't recommend one over the other. Any particular reason for your preferance?

prabhat00155: PEP8 doesn't recommend one over the other. Any particular reason for your preferance?

PR #76, opened 2019-03-08
Fixes #75, rename convert_sklearn into to_onnx, better pep8 for tests and documentation

### LINKED ISSUE #75: Rename convert_sklearn into to_onnx, deprecate convert_sklearn
To have a less ambiguous name: convert is a ambiguous as it does not precise in which way. to_onnx would precise a way and the format it is exported to.

xadupre:


docs/api_summary.rst
```diff
@@ -12,7 +12,7 @@ in *scikit-onnx*.
 Converters
 ==========
 
-.. autofunction:: skl2onnx.convert_sklearn
+.. autofunction:: skl2onnx.to_onnx
```
wenbingl: it would be better to discuss with PM to have a better name. In my opinion, convert_ and to_ actually I didn't see too much difference.

docs/examples/plot_complex_pipeline.py
```diff
@@ -100,36 +101,38 @@ def convert_dataframe_schema(df, drop=None):
...
+# Merging single column into vectors is not
 # the most efficient way to compute the prediction.
 # It could be done before converting the pipeline into a graph.
 
 ##################################
 # Convert the pipeline into ONNX
 # ++++++++++++++++++++++++++++++
 
-from skl2onnx import convert_sklearn
+from skl2onnx import to_onnx
```
wenbingl: A safer way is to keep the convert_sklearn and add a deprecated warning.

sdpython: This is what I did. But I replaced convert_sklearn by to_onnx everywhere.

xadupre: This is what I did but it is better to use the version without a warning.

docs/api_summary.rst
xadupre: Some people find this name confusing: they wonder "convert but in which way".

WilliamTambellini: I agree with @xadupre ...

c-bata:
Any progress on this?

docs/conf.py
```diff
@@ -5,13 +5,7 @@
 #
 # Configuration file for the Sphinx documentation builder.
 
-import os
-import sys
-import shutil
```
c-bata: These changes are conflicts with #100. If you want to merge this before #100, please tell me 👍 
https://github.com/onnx/sklearn-onnx/pull/100/files#diff-85987f48f1258d9ee486e3191495582dL10

xadupre: That's ok. I'll probably give up on this one as it does not seem to reach a consensus.

docs/api_summary.rst
xadupre: I drop this PR. It is easier to start from scratch than fixing conflicts. There is no point in renaming if it is not quickly merged.

PR #77, opened 2019-03-08
[WIP] test on scikit-learn 0.19.2

xadupre:


xadupre:
This will not pass the unit test unless we make some changes to the code to support 0.19.

prabhat00155:
Do you know how big are those changes going to be?

xadupre:
I would disable every model not supported with the current implementation (raise of an excetion) and disable all associated unit tests. That's the simplest.

xadupre:
Fixed in #93.

PR #79, opened 2019-03-10
Fixes for scikit-learn 0.21 + disable test tokenizing empty string with onnxruntime

xadupre:


tests/test_SklearnTfidfVectorizerConverter.py
```diff
@@ -40,8 +40,9 @@ def test_model_tfidf_vectorizer11_empty_string(self):
         model_onnx = convert_sklearn(vect, 'TfidfVectorizer',
                                      [('input', StringTensorType([1, 1]))])
         self.assertTrue(model_onnx is not None)
+        # onnxruntime fails with empty strings
```
wenbingl: Is it an issue of ONNXRuntime, if so, please file an issue for them in the GitHub.

xadupre: The unit test are passing right now which means it fails with onnxruntime 0.2.1 but fails with onnxruntime 0.3.0. However, I'm puzzled by this issue. I checked that onnxruntime had a test checking that an empty string would work. So I wanted to raise a more precise issue.

PR #81, opened 2019-03-12
Added MLPClassifier converter

prabhat00155:


tests/test_SklearnMLPConverter.py
```diff
@@ -0,0 +1,57 @@
...
+from sklearn.neural_network import MLPClassifier, MLPRegressor
+
+from skl2onnx import convert_sklearn
+from skl2onnx.common.data_types import FloatTensorType
+from test_utils import dump_data_and_model
+
+
+class TestSklearnCalibratedClassifierCVConverters(unittest.TestCase):
+
+    def test_model_mlp_classifier_binary(self):
```
xadupre: you should have test for every supported activation function

{‘identity’, ‘logistic’, ‘tanh’, ‘relu’}, default ‘relu’

prabhat00155: Makes sense, I'll add them.

prabhat00155: Added.

PR #84, opened 2019-03-13
Fix #82, #83, update the tokenizer, implements analyser=='char'

### LINKED ISSUE #82: Support analyzer='char' for the TfidfVectorizer
Currently the converter only supports analyzer='word' mode for the TfidfVectorizer, please add the 'char' mode support.

xadupre:


skl2onnx/operator_converters/TextVectorizer.py
```diff
@@ -46,16 +75,43 @@ def convert_sklearn_text_vectorizer(scope, operator, container):
...
+                raise NotImplementedError(
+                    "Only the default tokenizer based on default regular expression "
+                    "'{0}' is implemented.".format(default_pattern))
+        elif options['regex'] is not None:
+            if options['regex']:
+                regex = options['regex']
+            else:
+                regex = op.token_pattern
+                if regex == default_pattern:
+                    regex = '\\\\b(\\\\w\\\\w+)\\\\b'
```
xadupre: Using \\ instead of \ is unexpected but otherwise onnxruntime fails. It needs to be investigated.

skl2onnx/operator_converters/TextVectorizer.py
```diff
@@ -46,16 +75,43 @@ def convert_sklearn_text_vectorizer(scope, operator, container):
...
+        else:
+            regex = None
+            default_separators = options['sep']
+    else:
+        if options['sep'] != 'DEFAULT':
+            raise RuntimeError("Option sep has not effect if analyser != 'word'.")
+        if options['regex']:
+            regex = options['regex']
+        else:
+            regex = '[^\\\\n]'
```
xadupre: onnxruntime fails if this expression is replaced by '.' because of the same error that happens for empty strings.

skl2onnx/operator_converters/TextVectorizer.py
```diff
@@ -8,6 +8,51 @@
...
+    or ``('a', ' b')``.
+    See `ngram sequence
+    <https://github.com/scikit-learn/scikit-learn/blob/master/
+    sklearn/feature_extraction/text.py#L169>`_.
+    """
+    if op.analyzer == 'word':
+        if op.ngram_range[0] == op.ngram_range[1] == 1:
+            spl = [text]
+        elif '  ' not in text:
+            if text.startswith(' '):
```
prabhat00155: Line 25 to line 33 could just be replaced with:
spl = text.strip().split()

skl2onnx/operator_converters/TextVectorizer.py
```diff
@@ -46,16 +120,45 @@ def convert_sklearn_text_vectorizer(scope, operator, container):
...
+                    regex = '[a-zA-Z0-9_]+'
+            default_separators = None
+        else:
+            regex = None
+            default_separators = options['sep']
+    else:
+        if options['sep'] != 'DEFAULT':
+            raise RuntimeError("Option sep has not effect "
+                               "if analyser != 'word'.")
+        if options['regex']:
```
prabhat00155: There 4 lines could be shortened to:
regex = options['regex'] if options['regex'] else '.'
(more Pythonic)

tests/test_SklearnCountVectorizerConverter.py
```diff
@@ -1,15 +1,21 @@
 """
-Tests scikit-learn's binarizer converter.
+Tests scikit-learn's tfidf converter.
```
prabhat00155: Isn't this count vectorizer converter's test?

tests/test_SklearnTfidfVectorizerConverterChar.py
```diff
@@ -0,0 +1,122 @@
+"""
+Tests scikit-learn's binarizer converter.
```
prabhat00155: tfidf not binarizer.

skl2onnx/operator_converters/TextVectorizer.py
xadupre: Not exactly because I need to remove only the first space. If there are two spaces at the beginning, I'll only remove one.

tests/test_SklearnCountVectorizerConverter.py
xadupre: Fixed.

tests/test_SklearnTfidfVectorizerConverterChar.py
xadupre: Fixed. I think we should remove those comments. Errors due to copy paste and comments does not bring information.

skl2onnx/operator_converters/TextVectorizer.py
prabhat00155: But then you have a check for two spaces, which would fail and this wouldn't be executed in case there are two spaces.  
elif '  ' not in text:

xadupre: You are right, I also need to modify the way it is split: only one space should be used.

skl2onnx/operator_converters/TextVectorizer.py
xadupre: Fixed but I'd like flake8 to capture this otherwise let's leave it that way.

skl2onnx/operator_converters/TextVectorizer.py
xadupre: In fact, I changed the way I do. This code is only used when space can be part of tokens. There are some ambiguities we can clear but not all of them.

skl2onnx/operator_converters/TextVectorizer.py
xadupre: Fixed.

skl2onnx/operator_converters/TextVectorizer.py
xadupre: Fixed.

PR #85, opened 2019-03-13
Added support for weights and metric hyper-parameters in KNeightbours…

prabhat00155:
…Regressor converter


xadupre:
unit test is missing?

PR #88, opened 2019-03-15
upgrade version number

xadupre:


setup.py
```diff
@@ -32,7 +32,11 @@
     start_pos = long_description.find('## Introduction')
     if start_pos >= 0:
         long_description = long_description[start_pos:]
-
+    try:
```
wenbingl: I never encounter any text encoding with this script, Did build your wheel under Linux?

skl2onnx/__init__.py
```diff
@@ -7,7 +7,7 @@
 """
 Main entry point to the converter from the *scikit-learn* to *onnx*.
 """
-__version__ = "1.4.3"
+__version__ = "1.4.4.2"
```
wenbingl: Most python package version are all three parts, why do you want to suggest a different form?

xadupre: Sorry for that, I tried a couple of times to fix the rendering of README.md on pypi. I had to convert markdown into rst with pypandoc.

setup.py
xadupre: No, with windows.

wenbingl: In https://packaging.python.org/guides/making-a-pypi-friendly-readme/, it mentioned: 
setuptools >= 38.6.0
wheel >= 0.31.0
twine >= 1.11.0

Did your environment satisfy this requirements?

xadupre:
You can check the results of the new README.rst here: https://github.com/xadupre/sklearn-onnx/tree/master

setup.py
xadupre: I did not use twine, maybe that's why. I used python setup.py bdist_wheel upload.

skl2onnx/__init__.py
xadupre: I'll close the issue. The package must be uploaded with twine.

PR #90, opened 2019-03-19
[WIP] Fixes converters for scikit-learn 0.21.

xadupre:


setup.py
```diff
@@ -26,7 +26,7 @@
     if len(line) > 0:
         version_str = line[0].split('=')[1].strip('" ')
 
-README = os.path.join(os.getcwd(), "README.md")
+README = os.path.join(os.getcwd(), "README.rst")
```
wenbingl: It was request we should merge two README together. Instead of creating another one, you need figure it out why markdown doesn't work in your env. which is officially supported in the python doc.

xadupre: I need to rebase my branch as this is unrelated to the purpose of the PR. I'm currently trying to remove all the issues. I added [WIP].

xadupre:
Switched to another PR.

PR #98, opened 2019-04-01
Added int as allowed input type in TruncatedSVD and PCA converter

prabhat00155:


skl2onnx/operator_converters/SVD.py
```diff
@@ -22,10 +24,18 @@ def convert_truncated_svd(scope, operator, container):
         transform_matrix_name, onnx_proto.TensorProto.FLOAT,
         transform_matrix.shape, transform_matrix.flatten())
 
+    input_name = operator.inputs[0].full_name
+    if isinstance(operator.inputs[0].type, Int64TensorType):
+        cast_output_name = scope.get_unique_variable_name('cast_output')
+
+        apply_cast(scope, input_name, cast_output_name, container,
```
wenbingl: No need to cast it back to integer64?

prabhat00155: The output will always be float.

PR #99, opened 2019-04-04
Refactor common code (onnx_ops, data_types), include onnxconverter_common

vinitra:
- [x] add onnxconverter_common (pypi) to the linux CI build
- [x] add onnxconverter_common (pypi) to the windows CI build
- [x] add onnxconverter-common dependency to requirements.txt
- [x] import common ops from onnxconverter_common, merge changes from skl2onnx's onnx_ops.py
- [x] add wrappers for onnxconverter_common in skl2onnx common files


wenbingl:
Do we need to publish the common package firstly? Then it looks more easier to add this dependency for skl2onnx.

vinitra:
Yes, I think that would make the most sense @wenbingl!

.azure-pipelines/linux-conda-CI.yml
```diff
@@ -60,16 +60,20 @@ jobs:
...
       conda install protobuf
-      python -m pip install --upgrade pip      
+      python -m pip install --upgrade pip
     displayName: 'Install environment'
 
   - script: |
       pip install flake8
     displayName: 'install flake8'
 
+  - script: |
```
xadupre: I removed it for linux and let it for Windows. That way, a build does not get stuck because of style. It usually saves one commit.

.azure-pipelines/linux-conda-CI.yml
```diff
@@ -107,6 +111,13 @@ jobs:
       pip install pytest
     displayName: 'install requirements'
 
+  - script: |
```
xadupre: It would be better to take it from pypi because that's a version users will probably use.

skl2onnx/common/_container.py
```diff
@@ -34,8 +35,8 @@ def _get_operation_list():
                     found = g.groups()[0]
                     break
             if found is None:
-                raise RuntimeError("Unable to find an ONNX name in function "
-                                   "'{0}', source=\n{1}".format(k, source))
+                warnings.warn("Unable to find an ONNX name in function "
```
xadupre: It is needed?

skl2onnx/common/utils.py
```diff
@@ -4,234 +4,11 @@
 # license information.
 # --------------------------------------------------------------------------
 
-import numbers
+from onnxconverter_common.utils import *  # noqa
```
xadupre: It is not recommended to do that. The import never fails even if a function changed name.

.azure-pipelines/linux-conda-CI.yml
vinitra: Makes sense! I'll revert my change.

.azure-pipelines/linux-conda-CI.yml
vinitra: Yes, that would be preferable. It's not published yet, so we're waiting for that first.

skl2onnx/common/utils.py
vinitra: Will replace with function names.

skl2onnx/common/_container.py
vinitra: I talked to @wenbingl about this earlier -- a number of the operations in onnxconverter_common.onnx_ops are not being picked up by the regex command. Is there a better way to identify the operations without modifying the apply_op code for container.add_node('Name'...)? I think keras-onnx manually lists all the operators.

vinitra: (But yes, it is needed).

.azure-pipelines/linux-conda-CI.yml
vinitra: Resolved in latest update!

PR #103, opened 2019-04-10
Run flake8 in linux-conda-ci

c-bata:
Currently, flake8 is installed but not executed.

### DIFF
--- a/.azure-pipelines/linux-conda-CI.yml
+++ b/.azure-pipelines/linux-conda-CI.yml
@@ -71,6 +71,10 @@ jobs:
     displayName: 'install flake8'
 
   - script: |
+      flake8 skl2onnx
+    displayName: 'flake8'
+
+  - script: |
       pip install numpy$(numpy.version)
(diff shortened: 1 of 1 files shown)


.azure-pipelines/linux-conda-CI.yml
```diff
@@ -70,6 +70,10 @@ jobs:
       pip install flake8
     displayName: 'install flake8'
 
+  - script: |
```
c-bata: https://github.com/onnx/sklearn-onnx/blob/fd48cc12a2485f46ef4370e1473419fb219d21a8/.azure-pipelines/win32-conda-CI.yml#L52-L54

vinitra: Thanks for the PR, @c-bata! We just discussed this in #99 and decided to remove the install since flake8 already runs in the windows build.

cc: @xadupre

c-bata: I see. I close this.

PR #105, opened 2019-04-11
Fixed bug whereby sometimes RandomForestClassifier.n_estimators != le…

jim-meyer:
Fixed bug whereby sometimes RandomForestClassifier.n_estimators != len(RandomForestClassifier.estimators_) thereby causing an index error at conversion time for models like:
    classifier = RandomForestClassifier(n_estimators = 30, ...)
    for j in range(10):
        classifier.fit(X, y)
        classifier.n_estimators += 30

At this point classifier.n_estimators is 330 but len(classifier.estimators_) is 300. Thus conversion was failing on the non-existent 301st tree with an index error.


xadupre:
Thanks for fixing this bug.

skl2onnx/operator_converters/RandomForest.py
```diff
@@ -13,6 +13,17 @@
 from ..common.tree_ensemble import get_default_tree_regressor_attribute_pairs
 
 
+def _num_estimators(op):
```
prabhat00155: Why do you need a separate function for this?

tests/test_SklearnRandomForestConverters.py
```diff
@@ -30,11 +31,37 @@ def test_random_forest_classifier(self):
                                      allow_failure="StrictVersion(onnx.__version__) < StrictVersion('1.2') or "
                                                    "StrictVersion(onnxruntime.__version__) <= StrictVersion('0.2.1')")
 
+    @unittest.skipIf(not onnx_built_with_ml(),
+                     reason="Requires ONNX-ML extension.")
+    def test_random_forest_classifier_mismatched_estimator_counts(self):
```
prabhat00155: Could you add some elaborate tests where you can verify if the converter works fine? Look at this: https://github.com/onnx/sklearn-onnx/blob/master/tests/test_SklearnMLPConverter.py

tests/test_SklearnRandomForestConverters.py
```diff
@@ -5,13 +5,14 @@
...
+import numpy
 from sklearn.ensemble import RandomForestClassifier
 from sklearn.ensemble import RandomForestRegressor
 from sklearn.ensemble import ExtraTreesClassifier
 from sklearn.ensemble import ExtraTreesRegressor
-from skl2onnx.common.data_types import onnx_built_with_ml
+from skl2onnx.common.data_types import onnx_built_with_ml, FloatTensorType
 from test_utils import dump_one_class_classification, dump_binary_classification, dump_multiple_classification
-from test_utils import dump_multiple_regression, dump_single_regression
+from test_utils import dump_multiple_regression, dump_single_regression, convert_model, dump_data_and_model
```
prabhat00155: We don't have pep8 check on tests folder yet, still I would recommend running pep8 check locally using flake8. This line is clearly > 79 characters.

tests/test_utils/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 from .tests_helper import dump_data_and_model
 from .tests_helper import dump_one_class_classification, dump_binary_classification, dump_multiple_classification
-from .tests_helper import dump_multiple_regression, dump_single_regression
+from .tests_helper import dump_multiple_regression, dump_single_regression, convert_model
```
c-bata: Hi @jim-meyer! It looks this change isn't needed, right?

skl2onnx/operator_converters/RandomForest.py
jim-meyer: My thinking goes like this... I like to comment key decisions like this for the next person. Do I put the same comment in two places? Or "factor" the comment out to a separate function? I tend to take the latter for maintenance reasons in projects I work on though I recognize that is a stylistic decision. I'm ok changing this if that is what you strongly prefer.

tests/test_SklearnRandomForestConverters.py
jim-meyer: I will change this line to follow pep8. I saw the preceding line was even longer so I didn't think twice about adding to this line. Will be nice if we can get the unit tests all to pep8 and start enforcing on them too. I see https://github.com/onnx/sklearn-onnx/pull/76 has done just that among other things.

tests/test_utils/__init__.py
jim-meyer: Hey @c-bata , I'm not seeing the conflict with your changes yet. Can you elaborate? I had to add convert_model here since it is needed and adding it this seemed best to follow how dump_multiple_regression, dump_single_regression are imported. I saw how they worked and since my change needed to do something similar I just followed their lead. Or maybe you can suggest an alternative that is better?

tests/test_SklearnRandomForestConverters.py
jim-meyer: Hi @prabhat00155 , is that a general request for what I could do next to help? Or is not having more elaborate tests blocking your approval for this particular PR for some reason?

tests/test_utils/__init__.py
jim-meyer: Nvm, now I see the conflict after I refresh from master.  Let me grab the latest and I'll update it to be compatible with the latest code. Stay tuned.

skl2onnx/operator_converters/RandomForest.py
prabhat00155: I am not sure if you need the comments here. In case you feel it's necessary, then yes this seems all right.

tests/test_SklearnRandomForestConverters.py
prabhat00155: I just feel that would be a more comprehensive test. I don't think fitting on 3 data points and predicting on the same checks the correctness of the converter properly. Anyways, I think you are mainly focussed on testing your fix, so I'll approve your PR.

PR #109, opened 2019-04-17
Add onnxruntime nightly build to CI

xadupre:


wenbingl:
Should we consider seperate the nightly build verification from CI build? otherwise, any issue in onnxruntime will block all PR for skl2onnx converter. @vinitra for onnxmltools as well

xadupre:
It would be better to have a separate build, it should run a regular basis and not necessarily at every commit.

vinitra:
For onnxmltools, I created a separate nightly build to validate the onnxruntime nightly test PyPi release with two new pipeline files -- 1 for windows, 1 for linux. See https://github.com/onnx/onnxmltools/pull/284

.azure-pipelines/linux-conda-CI.yml
```diff
@@ -107,6 +118,15 @@ jobs:
...
 
+  - script: |
+      if [ '$(onnxrt.version)' == 'nightly' ]
+      then
+        pip install -i https://test.pypi.org/simple/ ort-nightly
+      else
+        pip install onnxruntime$(onnxrt.version)
+      fi
+    displayName: 'install onnxruntime again (previous step might remove it)'
+
```
vinitra: Why is this needed twice? If you do this after requirements-dev.txt, then it only needs to be done once.

.azure-pipelines/linux-conda-CI.yml
```diff
@@ -19,6 +19,12 @@ jobs:
         onnx.version: 'git'
         onnxrt.version: '==0.2.1'
         sklearn.version: '==0.20.3'
+      Py37-OnnxGit-Nightly:
+        python.version: '3.7'
+        numpy.version: '==1.16.2'
+        onnx.version: 'git'
+        onnxrt.version: 'nightly'
+        sklearn.version: '==0.20.3'
```
vinitra: Make a separate file for just the nightly release check, and then delete all other python build versions from that file. See onnxmltools [linux-CI-nightly file](https://github.com/onnx/onnxmltools/pull/284/files#diff-6989b99615ec0b18248271b18aac06e2) for reference.

xadupre:
Thanks for the tip. I'm testing Something as the build on linux was unexpectedly succeeding. I noticed that pip install requiremens-dev reinstalled onnxruntime despite the fact it was already installed but only on linux. It does not happen on Windows. I was puzzled (I'm still puzzled). I'll fix that today.

xadupre:
CI would pass with #84.

.azure-pipelines/linux-conda-CI.yml
xadupre: I removed. It did not have any impact. I still get onnxruntime: 0.3.0 on linux and onnxruntime: 0.3.1 on windows.

vinitra:
Closing in favor of #114

PR #111, opened 2019-04-24
Enables opset 9 for CountVectorizer and TfidfVectorizer

xadupre:


skl2onnx/operator_converters/TextVectorizer.py
```diff
@@ -52,7 +52,7 @@ def _intelligent_split(text, op, tokenizer, existing):
     if spl in existing:
         raise RuntimeError("The converter cannot guess how to "
                            "split an expression into tokens.")
-    if op.ngram_range[0] == 1:
+    if op.ngram_range[0] == 1 and op.ngram_range[1] > 1:
```
wenbingl: do we need to check len(op.ngram_range) > 1 before check the its value?

tests/test_SklearnCountVectorizerConverterBug.py
```diff
@@ -0,0 +1,52 @@
...
+from skl2onnx import convert_sklearn
+from skl2onnx.common.data_types import StringTensorType
+import onnx
+from test_utils import dump_data_and_model
+
+
+class TestSklearnCountVectorizerBug(unittest.TestCase):
+
+    @unittest.skipIf(
+        StrictVersion(onnx.__version__) <= StrictVersion("1.4.1"),
```
wenbingl: less and equal?

skl2onnx/operator_converters/TextVectorizer.py
prabhat00155: len(ngram_range) will always be 2 in a fit model.

PR #117, opened 2019-05-02
update CI to onnxruntime 0.4, onnx 1.5

xadupre:


tests/test_SklearnTfidfVectorizerConverter.py
```diff
@@ -292,13 +292,15 @@ def test_model_tfidf_vectorizer11parenthesis_class(self):
...
         )
         self.assertTrue(model_onnx is not None)
+        # This test depends on this issue:
+        # https://github.com/Microsoft/onnxruntime/issues/957.
         dump_data_and_model(
             corpus,
             vect,
             model_onnx,
             basename="SklearnTfidfVectorizer11ParenthesisClass-OneOff-SklCol",
             allow_failure="StrictVersion(onnxruntime.__version__)"
```
wenbingl: the latest onnxruntime is 0.4.0, is it intented to ignore the failure as well. If that case, can we just disable this case?

xadupre: It depends whether or not onnxruntime decides to support this case. It used to pass. I let the test just in case.

PR #121, opened 2019-05-03
Fixed probability calculation in SGDClassifier

prabhat00155:


skl2onnx/operator_converters/LinearClassifier.py
```diff
@@ -56,7 +56,7 @@ def convert_sklearn_linear_classifier(scope, operator, container):
                 classifier_attrs['post_transform'] = 'LOGISTIC'
         else:
             classifier_attrs['post_transform'] = 'LOGISTIC'
-    elif op.__class__.__name__ in ('SGDClassifier', 'LinearSVC'):
+    elif op.__class__.__name__ in ('LinearSVC'):
```
wenbingl: type(op) ?

prabhat00155: I just removed SGDClassifier from here. Anyways, type(op) can't directly be used here as that would return the type of object(something like sklearn.svm.classes.LinearSVC) and here it is being compared to the string itself.

PR #122, opened 2019-05-06
First tools to inspect intermediate nodes of a scikit-learn pipeline and their onnx conversion

xadupre:


skl2onnx/convert.py
```diff
@@ -123,4 +126,4 @@ def convert_sklearn(model, name=None, initial_types=None, doc_string='',
     onnx_model = convert_topology(topology, name, doc_string, target_opset,
                                   options=options)
 
-    return onnx_model
+    return (onnx_model, topology) if intermediate else onnx_model
```
wenbingl: can we make this intermediate stuff to a global object which the user can set flag and get the data from it so we don't need alter the API?
BTW, the topology is not a class we want to expose as API, it's better to wrap it as func in a more standard way.

xadupre: I really don't like global objects. It is hard to debug and global variable it usually a bad design. topology is not exposed but azureml.automl is using it. I don't specifically need topology, my goal is to map ONNX nodes scikit-learn operators and the topology has that information. I can extract it and return it instead of the topology but it seems more simple this way.

docs/examples/plot_investigate_pipeline.py
```diff
@@ -0,0 +1,132 @@
...
+import matplotlib.pyplot as plt
+import pandas as pd
+
+from sklearn import datasets
+from sklearn.decomposition import PCA
+from sklearn.linear_model import LogisticRegression
+from sklearn.pipeline import Pipeline
+from sklearn.model_selection import GridSearchCV
+
+logistic = LogisticRegression()
```
prabhat00155: What's the need to declare these two variables?

docs/examples/plot_investigate_pipeline.py
```diff
@@ -0,0 +1,132 @@
...
+
+import numpy, sklearn
+print("numpy:", numpy.__version__)
+print("scikit-learn:", sklearn.__version__)
+import onnx, onnxruntime, skl2onnx, onnxmltools, lightgbm
+print("onnx: ", onnx.__version__)
+print("onnxruntime: ", onnxruntime.__version__)
+print("skl2onnx: ", skl2onnx.__version__)
+
+
```
prabhat00155: Why do you have newlines at the end of this file?

skl2onnx/helpers/intermediate.py
```diff
@@ -0,0 +1,171 @@
...
+                rows.append('  )')
+            else:
+                raise KeyError(
+                    "Unable to find output for method '{}'.".format(k))
+        return "\n".join(rows)
+
+    def display(self, data, nrows):
+        """
+        Displays the first
+        """
```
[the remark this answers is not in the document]
xadupre: They really have a rule for everything. We should enforce that in CI.

docs/examples/plot_investigate_pipeline.py
xadupre: Fixed.

PR #129, opened 2019-05-10
Fix GradientBoostedTrees

xadupre:


tests/test_SklearnGradientBoostingConverters.py
```diff
@@ -6,22 +6,49 @@
...
+        for cl in [1e-6, 0.9, 0.5, 0., 1e-5]:
+            model.init_.class_prior_ = np.array([cl, cl])
+            initial_types = [('input', FloatTensorType((1, X.shape[1])))]
+            model_onnx = convert_sklearn(model, initial_types=initial_types)
+            if "Regressor" in str(model_onnx):
+                raise AssertionError(str(model_onnx))
+            sess = InferenceSession(model_onnx.SerializeToString())
+            res = sess.run(None, {'input': X.astype(np.float32)})
+            pred = model.predict_proba(X)
+            if res[1][0][0] != pred[0, 0]:
```
prabhat00155: Why are you not using dump_data_and_model() here?

tests/test_utils/tests_helper.py
```diff
@@ -442,6 +455,28 @@ def dump_multiple_classification(
...
+                                       [("input", FloatTensorType([1, 2]))])
+    if verbose:
+        print("[dump_multiple_classification] model was converted")
+    dump_data_and_model(
+        X.astype(numpy.float32),
+        model,
+        model_onnx,
+        folder=folder,
+        allow_failure=allow_failure,
+        basename=prefix + "Mcl" + model.__class__.__name__ + suffix + 'Rnd',
```
prabhat00155: Instead of so many concatenations, f-strings would be better.

tests/test_utils/tests_helper.py
```diff
@@ -398,6 +396,22 @@ def dump_binary_classification(
...
+    model.fit(X, y)
+    model_onnx, prefix = convert_model(model, "binary classifier",
+                                       [("input", FloatTensorType([1, 2]))])
+    dump_data_and_model(
+        X.astype(numpy.float32),
+        model,
+        model_onnx,
+        folder=folder,
+        allow_failure=allow_failure,
+        basename=prefix + "Bin" + model.__class__.__name__ + suffix + 'Rnd',
```
prabhat00155: f-strings would be better.

tests/test_SklearnDecisionTreeConverters.py
```diff
@@ -5,18 +5,54 @@
...
+        X, y = make_classification(10, n_features=4, random_state=42)
+        X = X[:, :2]
+        model.fit(X, y)
+        initial_types = [('input', FloatTensorType((1, X.shape[1])))]
+        model_onnx = convert_sklearn(model, initial_types=initial_types)
+        sess = InferenceSession(model_onnx.SerializeToString())
+        res = sess.run(None, {'input': X.astype(np.float32)})
+        pred = model.predict_proba(X)
+        if res[1][0][0] != pred[0, 0]:
+            raise AssertionError("{}\n--\n{}".format(pred, DataFrame(res[1])))
```
prabhat00155: Any reason for not using dump_data_and_model()?

xadupre: For investigation, I need more details results.

tests/test_utils/tests_helper.py
xadupre: f-strings do not work in Python 3.5.

tests/test_utils/tests_helper.py
xadupre: See below.

tests/test_SklearnGradientBoostingConverters.py
xadupre: See below.

tests/test_SklearnDecisionTreeConverters.py
prabhat00155: If this is just for investigation, why are you checking it in?

xadupre: To run it on multiple devices. For some reasons, it works on Linux but not on Windows. I'll probably leave it that way because dump_model also dumps the model on disk and it is not needed for this test. It is faster this way.

prabhat00155: Okay

PR #132, opened 2019-05-13
Fixes #130, Imputer was removed in scikit-learn 0.21

### LINKED ISSUE #130: from sklearn.preprocessing import Imputer fails in scikit-learn 0.21

xadupre:


skl2onnx/_supported_operators.py
```diff
@@ -75,7 +75,11 @@
     # 0.19
     from sklearn.preprocessing import Imputer as SimpleImputer
 from sklearn.preprocessing import Binarizer
-from sklearn.preprocessing import Imputer
+try:
```
wenbingl: Can you add a explicit version checker here instead of try block, which seems more clear to the user and doesn't mix with the other issue?

xadupre: People usually do it this way: https://github.com/scikit-learn-contrib/sklearn-pandas/blob/master/sklearn_pandas/cross_validation.py#L2. Exception are cheap in python.

xadupre: Other comment: dealing with version number means also dealing with weird version number (.dev0, …) and it may introduce a bug if somebody use a weird version. I prefer to stick to exception.

wenbingl: It's your design decision, If you think after sklearn 0.21, Imputer will be not supported, the version checking looks better. If you think the converter still works well even without Imputer, try-block looks more robust. This is the straightforward idea when they read the code.

xadupre: Imputer = SimpleImputer, both existed until 0.21 removed Imputer. I like try catch otherwise the comparison will have to deal version such as x.x.dev0, x.x.rc1, x.x.beta… I would have to write a couple of unit test just to make sure the comparison works in all cases.

Issue #118, opened 2019-05-02
Update SVM converter for scikit-learn 0.21


xadupre:
This PR was merged: https://github.com/scikit-learn/scikit-learn/pull/10440 and has an impact on SVM. Issue raised here: https://github.com/scikit-learn/scikit-learn/issues/13030#event-2089752816.

PR #133, opened 2019-05-15
Addresses #118, fixes unit test for SVC when probabilities is False

xadupre:
When probability=False, onnxruntime and scikit-learn do not return the same raw score. scikit-learn returns aggregated score with a matrix [N, C] and onnxruntime returns the raw score from svm algorithm with a matrix [N, (C(C-1)/2].

### DIFF
--- a/tests/test_SklearnSVMConverters.py
+++ b/tests/test_SklearnSVMConverters.py
@@ -119,6 +119,8 @@ class TestSklearnSVM(unittest.TestCase):
             model,
             model_onnx,
             basename="SklearnBinSVCLinearPF-NoProbOpp",
+            allow_failure="StrictVersion(onnxruntime.__version__)"
+                          " <= StrictVersion('0.5.0')"
         )
 
     def test_convert_svc_binary_linear_ptrue(self):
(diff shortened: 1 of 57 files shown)


tests/test_SklearnSVMConverters.py
```diff
@@ -78,13 +118,70 @@ def test_convert_svmc_linear_binary(self):
             X,
             model,
             model_onnx,
-            basename="SklearnBinSVCLinearPF",
-            allow_failure=True,
+            basename="SklearnBinSVCLinearPF-NoProbOpp",
+            allow_failure="StrictVersion(onnxruntime.__version__)"
+                          " <= StrictVersion('0.5.0')"
```
vinitra: Will the changes be made in ONNX Runtime by the version 0.5.0 release? Perhaps < is more appropriate that <=

xadupre: Good call :), the PR is already in onnxruntime so < is better.

tests/test_SklearnSVMConverters.py
```diff
@@ -78,13 +118,70 @@ def test_convert_svmc_linear_binary(self):
...
             model,
             model_onnx,
-            basename="SklearnBinSVCLinearPF",
-            allow_failure=True,
+            basename="SklearnBinSVCLinearPF-NoProbOpp",
+            allow_failure="StrictVersion(onnxruntime.__version__)"
+                          " < StrictVersion('0.5.0')"
         )
 
-    def test_convert_svmc_linear_multi(self):
```
vinitra: Why were these originally named "svmc" before? I noticed the same naming structure in onnxmltools' unit tests for libSVM.

xadupre: Probably me, I changed for something more explicit.

PR #139, opened 2019-05-21
Handle scikit-learn dev version when using StrictVersion in tests

xadupre:


tests/test_SklearnGLMClassifierConverter.py
```diff
@@ -11,6 +11,12 @@
 from test_utils import dump_data_and_model
 
 
+def sklearn_version():
```
prabhat00155: Do you want to name this _sklearn_version() as it's an internal function?

xadupre: Fixed.

PR #150, opened 2019-05-27
Fixes #147, fix probabilities when multi_class == 'multinomial'

xadupre:

PR #151, opened 2019-05-28
Fixed Logistic Regression converter

prabhat00155:


skl2onnx/operator_converters/logistic_regression.py
```diff
@@ -0,0 +1,88 @@
+# -------------------------------------------------------------------------
```
xadupre: I'm wondering why you need to create a separate converter instead of fixing the current one which is the direction I followed in the other PR #150.

prabhat00155: I think we can go ahead with your PR. For some reason I saw some issues when I tried to fix the linear_classifier converter.

xadupre: I let you review it then.

PR #154, opened 2019-05-30
Support int features in linear regressor converter(s)

prabhat00155:


skl2onnx/operator_converters/linear_regressor.py
```diff
@@ -18,7 +20,12 @@ def convert_sklearn_linear_regressor(scope, operator, container):
                            else [float(op.intercept_)])
     if len(op.coef_.shape) == 2:
         attrs['targets'] = op.coef_.shape[0]
-    container.add_node(op_type, operator.input_full_names,
+
+    cast_input_name = scope.get_unique_variable_name('cast_input')
+
+    apply_cast(scope, operator.input_full_names, cast_input_name,
```
xadupre: I assume this node is removed by the optimizer if the cast does not change the type. Maybe we should add a unit test to check the node disappears if that's the case.

prabhat00155: Added a check to only apply cast when input is of type int.

PR #157, opened 2019-05-31
Support int feature type in non-linear regressor converters

prabhat00155:


skl2onnx/operator_converters/gradient_boosting.py
```diff
@@ -103,7 +105,12 @@ def convert_sklearn_gradient_boosting_regressor(scope, operator, container):
         add_tree_to_attribute_pairs(attrs, False, tree, tree_id, tree_weight,
                                     0, False)
 
-    container.add_node(op_type, operator.input_full_names,
+    cast_input_name = scope.get_unique_variable_name('cast_input')
+
+    apply_cast(scope, operator.input_full_names, cast_input_name,
```
xadupre: I don't see the corresponding unit test. We should also have one to make sure the cast node disappears if the optimizer tells if it is redundant.

prabhat00155: I have put in a check before casting for int input type.

PR #173, opened 2019-06-07
Review error message raised in exceptions

xadupre:


skl2onnx/common/_topology.py
```diff
@@ -464,7 +465,7 @@ def _generate_unique_name(seed, existing_names):
         :return: a string similar to the seed
         """
         if seed == '':
-            raise ValueError('Name seed must be an non-empty string')
+            raise ValueError('Name seed must be an non-empty string.')
```
prabhat00155: Typo: a non-empty instead of an non-empty.

PR #179, opened 2019-06-12
Test topology pruning

xadupre:


tests/test_topology_prune.py
```diff
@@ -0,0 +1,77 @@
+"""
+Tests scikit-learn's binarizer converter.
```
prabhat00155: binarizer converter?

tests/test_topology_prune.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def __init__(self):
+        IdentityTransformer.__init__(self)
+
+
+def dummy_shape_calculator(operator):
+    op_input = operator.inputs[0]
+    N = op_input.type.shape[0]
+    C = op_input.type.shape[1]
+    operator.outputs[0].type = FloatTensorType([N, C])
```
prabhat00155: Do we really need N and C here? We could just write: 
operator.outputs[0].type = FloatTensorType(op_input.type.shape), couldn't we?

prabhat00155:
Why do you have so many commits in the PR, may be clean up your branch?

tests/test_topology_prune.py
xadupre: I don't know. I update the master branch on my repo and I create new branches with git checkout -b …

prabhat00155: You fork is not synchronised with master. Look at this: https://stackoverflow.com/questions/41283955/github-keeps-saying-this-branch-is-x-commits-ahead-y-commits-behind

PR #187, opened 2019-06-18
Fix missing node names

xadupre:
No unit test is capturing that it is missing. I wonder whether or not it is part of the spec that a node name cannot be empty. If so, we should assert in container if that happens.


wenbingl:
> No unit test is capturing that it is missing. I wonder whether or not it is part of the spec that a node name cannot be empty. If so, we should assert in container if that happens.

the node name is not mandatory, while it will be a good practice to have it

tests/test_SklearnGradientBoostingConverters.py
```diff
@@ -0,0 +1,113 @@
+# -------------------------------------------------------------------------
```
prabhat00155: Why have you added this file? We already have https://github.com/onnx/sklearn-onnx/blob/master/tests/test_sklearn_gradient_boosting_converters.py, with the unit tests.

xadupre: Sorry wrong rebase.

PR #188, opened 2019-06-20
Extend coverage for OneHotEncoder

xadupre:


tests/test_sklearn_one_hot_encoder_converter.py
```diff
@@ -155,6 +155,56 @@ def test_one_hot_encoder_one_string_one_int_cat(self):
...
 
+    @unittest.skipIf(
+        not one_hot_encoder_supports_string(),
+        reason="OneHotEncoder does not support this in 0.19",
+    )
+    def test_model_one_hot_encoder_list_sparse(self):
+        model = OneHotEncoder(categories=[[0, 1, 4, 5],
+                                          [1, 2, 3, 5],
+                                          [0, 3, 4, 6]],
+                              sparse=True)
```
prabhat00155: Can we return sparse matrix? How do you validate scikit and onnx outputs?

xadupre: No, everything is converted into dense.

prabhat00155: Then what's the point of this unit test? What does it check?

xadupre: I added it to have explicit tests for dense and sparse datasets when onnxruntime implements it.

PR #205, opened 2019-06-25
Improve OnnxOperator to support more types

xadupre:


skl2onnx/algebra/graph_state.py
```diff
@@ -70,14 +70,21 @@ def _add_constant(self, cst):
                 self.operator_name + 'cst')
             if cst.dtype in (np.float32, np.float64):
                 ty = onnx_proto.TensorProto.FLOAT
+                astype = np.float64
+            elif cst.dtype == np.int32:
+                ty = onnx_proto.TensorProto.INT32
```
prabhat00155: You are setting this to INT32 but typecasting to int64 in line 87. Is that what you intended to do, could you explain?

xadupre: Not sure if it is a bug or not but in onnx.helper.py, https://github.com/onnx/onnx/blob/master/onnx/helper.py#L231, the function check that number of float64 (not float32) so I'm just trying to avoid the exception at line https://github.com/onnx/onnx/blob/master/onnx/helper.py#L251.

PR #209, opened 2019-06-26
Fix SVC output

prabhat00155:


skl2onnx/_parse.py
```diff
@@ -252,6 +252,8 @@ def _parse_sklearn_column_transformer(scope, model, inputs,
 def _parse_sklearn_classifier(scope, model, inputs, custom_parsers=None):
     probability_tensor = _parse_sklearn_simple_model(
             scope, model, inputs, custom_parsers=custom_parsers)
+    if model.__class__ in [NuSVC, SVC] and not model.probability:
```
xadupre: No unit test?

prabhat00155: We already have unit tests in place: https://github.com/onnx/sklearn-onnx/blob/master/tests/test_sklearn_svm_converters.py

PR #213, opened 2019-06-27
Use n_estimators_ when set in GradientBoostingClassifier

prabhat00155:


skl2onnx/operator_converters/gradient_boosting.py
```diff
@@ -70,13 +70,15 @@ def convert_sklearn_gradient_boosting_classifier(scope, operator, container):
         raise ValueError('Labels must be all integer or all strings.')
 
     tree_weight = op.learning_rate
+    n_est = (op.n_estimators_ if hasattr(op, 'n_estimators_') else
```
xadupre: Why not len(estimators_)?

prabhat00155: It is same as n_estimators_.

PR #219, opened 2019-07-15
Fixed gradient boosting converters for different values of init hyper…

prabhat00155:
…-parameter

### DIFF
--- a/skl2onnx/operator_converters/gradient_boosting.py
+++ b/skl2onnx/operator_converters/gradient_boosting.py
@@ -28,32 +28,39 @@ def convert_sklearn_gradient_boosting_classifier(scope, operator, container):
     attrs = get_default_tree_classifier_attribute_pairs()
     attrs['name'] = scope.get_unique_operator_name(op_type)
 
-    if op.n_classes_ == 2:
-        transform = 'LOGISTIC'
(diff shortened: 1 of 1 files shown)


skl2onnx/operator_converters/gradient_boosting.py
```diff
@@ -28,32 +28,39 @@ def convert_sklearn_gradient_boosting_classifier(scope, operator, container):
...
+                base_values = op.init_.priors
     else:
-        transform = 'SOFTMAX'
-        # class_prior_ was introduced in scikit-learn 0.21.
-        if hasattr(op.init_, 'class_prior_'):
-            base_values = op.init_.class_prior_
-        else:
-            base_values = op.init_.priors
+        raise NotImplementedError(
+            'Setting init to an estimator is not supported, you may raise an '
```
wenbingl: If setting init is doable, probably you can add an issue by yourself and put the link the exception, is it better?

prabhat00155: It's not as straight-forward. If init is set to an estimator, we'll have to call it's predict(), which we could do but the result would be a set of output names(which will be evaluated by onnxruntime). However, TreeClassifier and TreeRegressor onnx ops expect a list of floats as base_values. This makes the implementation tricky.

PR #222, opened 2019-07-17
Handle transformer_weights hyper-parameter in FeatureUnion

prabhat00155:


skl2onnx/_parse.py
```diff
@@ -160,6 +160,18 @@ def _parse_sklearn_feature_union(scope, model, inputs, custom_parsers=None):
             _parse_sklearn_simple_model(
                 scope, transform, inputs,
                 custom_parsers=custom_parsers)[0])
+        if model.transformer_weights is not None:
+            transform_result = [transformed_result_names.pop()]
+            # Create a Multiply ONNX node
+            multiply_operator = scope.declare_local_operator('SklearnMultiply')
+            multiply_operator.inputs = transform_result
+            assert name in model.transformer_weights
```
xadupre: An explicit message would be better or remove it if it is always true.

skl2onnx/operator_converters/multiply_op.py
```diff
@@ -0,0 +1,23 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+from ..common._apply_operation import apply_mul
+from ..common._registration import register_converter
+from ..proto import onnx_proto
+
+
+def convert_sklearn_multiply(scope, operator, container):
```
xadupre: unit test?

prabhat00155: Adding them, will update my PR soon.

skl2onnx/_parse.py
prabhat00155: Actually scikit just keeps the actual value in this case, I'll change the behaviour to align with it.

PR #225, opened 2019-07-24
update for onnxruntime 0.5

xadupre:


skl2onnx/common/_topology.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, raw_name, onnx_name, scope, type=None):
                     raise TypeError("shape must be a tuple or a list not "
                                     "{}.".format(type_fct(shape)))
             for dim in shape:
-                if dim == 'None':
+                if dim in ('None', 'N', None):
```
prabhat00155: What do 'None' and 'N' mean?

skl2onnx/operator_converters/k_bins_discretiser.py
```diff
@@ -23,13 +26,7 @@ def convert_sklearn_k_bins_discretiser(scope, operator, container):
...
                       else [np.finfo(np.float32).max], op.bin_edges_))
     digitised_output_name = [None] * len(ranges)
-    instances = operator.inputs[0].type.shape[0]
-
-    last_column_name = scope.get_unique_variable_name('last_column')
-
-    container.add_initializer(
-        last_column_name, onnx_proto.TensorProto.FLOAT,
-        [instances, 1], np.ones(instances))
+    last_column_name = None
```
prabhat00155: Can you explain this change? You are setting last_column_name to None and then checking if last_column_name is None. I don't see any updates in between.

skl2onnx/operator_converters/label_binariser.py
```diff
@@ -14,10 +14,9 @@ def convert_sklearn_label_binariser(scope, operator, container):
     """Converts Scikit Label Binariser model to onnx format."""
     binariser_op = operator.raw_operator
     classes = binariser_op.classes_
-    input_samples = operator.inputs[0].type.shape[0]
-    zeros_tensor = np.full((input_samples, len(classes)),
+    zeros_tensor = np.full((1, len(classes)),
                            binariser_op.neg_label, dtype=np.float)
-    unit_tensor = np.full((input_samples, len(classes)),
+    unit_tensor = np.full((1, len(classes)),
```
prabhat00155: Why these changes, are we not planning on supporting batch prediction?

xadupre: Yes we do. It is using broadcasting. We cannot use input_samples in any converter because it is not known at conversion time. I replaced by 1 because we probably tested this converter either with the same number of samples or with we specifiied 1 input sample at conversion time.

skl2onnx/common/_topology.py
xadupre: 'N' is not needed anymore, I'll revert it. Wenbing made a change in onnxconverter-common, and I replaced 'N' by None everywhere and this means one dimension is unknown.

PR #238, opened 2019-08-27
Fixed NaN probability scores in BernoulliNB converter when alpha=0

prabhat00155:


skl2onnx/operator_converters/naive_bayes.py
```diff
@@ -237,8 +237,9 @@ def convert_sklearn_naive_bayes(scope, operator, container):
         dot_prod_name = scope.get_unique_variable_name('dot_prod')
         partial_sum_result_name = scope.get_unique_variable_name(
                                                 'partial_sum_result')
-
-        container.add_initializer(constant_name, proto_type, [], [1.0])
+        # Define constant slightly greater than 1 to avoid log 0
+        # scenarios when calculating log (1 - x) and x=1 in line 273
+        container.add_initializer(constant_name, proto_type, [], [1.000000001])
```
xadupre: Does it work? It think (float)1.000000001 == 1, can you tell more about the fix works?

prabhat00155: Yes, it works. I found that in the case where I saw NaN probability, np.exp(model.feature_log_prob_) had values of 1.00000000e+00. In the code we then subtract this value from constant_name and take a log, which would result in NaN. So, I changed the constant to slightly greater than 1 which fixes the issue.

prabhat00155: I checked the model, here is how the constant is represented in the model:
id: constant
kind: Initializer
type: float32
1.0000009536743164

prabhat00155: Before the fix, the constant looked like this in the model:
id: constant
kind: Initializer
type: float32
1

PR #241, opened 2019-09-02
Added missing pandas to requirements.txt

vriesdemichael:
Since pandas is used when importing convert_sklearn it should be added as requirement.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ protobuf
 onnx>=1.2.1
 scikit-learn>=0.19
 onnxconverter-common>=1.5.1
+pandas


requirements.txt
```diff
@@ -4,3 +4,4 @@ protobuf
 onnx>=1.2.1
 scikit-learn>=0.19
 onnxconverter-common>=1.5.1
+pandas
```
xadupre: pandas is used only in investigate.py, I would prefer to change the only that refers to it instead of adding a new dependency not really needed.

prabhat00155:
Closing this PR, as it has been inactive for more than 3 months.

PR #244, opened 2019-09-03
Fix the build.

xadupre:


tests/test_algebra_meta_onnx.py
```diff
@@ -174,6 +174,12 @@ def _check_algebra_onnxruntime(self, untested=None, model=None,
             if node.op_type in untested:
                 return (node.op_type, False,
                         "cannot load ONNX model {}".format(e))
+            if StrictVersion(onnx.__version__) >= StrictVersion("1.5.0"):
+                # onnx 1.5.0 and onnx dev version have the same version number
+                # some operators defined in onnx nut not yet implemented in
```
prabhat00155: typo? I think you meant to write 'but' instead of 'nut'.

xadupre: very likely :)

PR #246, opened 2019-09-04
Added GridSearchCV converter

prabhat00155:


skl2onnx/operator_converters/grid_search_cv.py
```diff
@@ -0,0 +1,38 @@
...
+    best_estimator = grid_search_op.best_estimator_
+    op_type = sklearn_operator_name_map[type(best_estimator)]
+    grid_search_operator = scope.declare_local_operator(op_type)
+    grid_search_operator.raw_operator = best_estimator
+    grid_search_operator.inputs = operator.inputs
+    label_name = scope.declare_local_variable('label')
+    grid_search_operator.outputs.append(label_name)
+    if type(best_estimator) in sklearn_classifier_list:
+        proba_name = scope.declare_local_variable('probability_tensor',
+                                                  FloatTensorType())
```
xadupre: You need to add a test with the GaussianProcess and the use of double to make sure this code passes.

skl2onnx/_parse.py
```diff
@@ -271,6 +274,14 @@ def _parse_sklearn_column_transformer(scope, model, inputs,
         return transformed_result_names
 
 
+def _parse_sklearn_grid_search_cv(scope, model, inputs, custom_parsers=None):
+    return (_parse_sklearn_classifier(
+                scope, model, inputs, custom_parsers=None)
+            if type(model.best_estimator_) in sklearn_classifier_list else
```
xadupre: Maybe use this function: https://scikit-learn.org/stable/modules/generated/sklearn.base.is_classifier.html ?

prabhat00155: I didn't know about this function, looks perfect in this scenario, thanks!

skl2onnx/operator_converters/grid_search_cv.py
prabhat00155: Added GaussianProcessRegressor unit tests with both float and double data types.

PR #253, opened 2019-09-11
Add check for converters that need decision_function()

prabhat00155:


skl2onnx/operator_converters/calibrated_classifier_cv.py
```diff
@@ -385,6 +386,14 @@ def convert_sklearn_calibrated_classifier_cv(scope, operator, container):
                               [], [clf_length])
 
     for clf in op.calibrated_classifiers_:
+        if (hasattr(clf.base_estimator, 'decision_function') and
```
xadupre: Not sure where the exception should be raised. Here or in the converter itself.

prabhat00155: Well, this issue only comes up in case of CalibratedClassifierCV currently, hence the expection here.

PR #261, opened 2019-09-18
Fixes VotingClassifier discrepencies + enable batch predictions

xadupre:


tests/test_sklearn_voting_classifier_converter.py
```diff
@@ -73,6 +75,8 @@ def test_operator_mul(self):
```
prabhat00155: Could you add unit tests that reflect the fix made, that would have failed otherwise?

skl2onnx/operator_converters/voting_classifier.py
```diff
@@ -48,56 +51,71 @@ def convert_voting_classifier(scope, operator, container):
...
-        container.add_node(
-            op_name, conc_name, red_name,
-            name=scope.get_unique_operator_name(op_name), axes=[0])
+        container.add_node('Sum', probs_names,
+                           operator.outputs[1].full_name,
+                           name=scope.get_unique_operator_name('Sum'))
     else:
-        raise NotImplementedError(
+        raise RuntimeError(
             "flatten_transform==True is not implemented yet. "
```
prabhat00155: Why RuntimeError rather than NotImplementedError?

xadupre: Fixed.

tests/test_sklearn_voting_classifier_converter.py
xadupre: Fixed.

PR #264, opened 2019-09-20
Add unit tests to run example in the documentation

xadupre:


skl2onnx/_parse.py
```diff
@@ -198,7 +198,9 @@ def _parse_sklearn_column_transformer(scope, model, inputs,
     # Output variable name of each transform. It's a list of string.
     transformed_result_names = []
     # Encode each transform as our IR object
-    for name, _, column_indices in model.transformers_:
+    for name, op, column_indices in model.transformers_:
+        if op == 'drop':
```
prabhat00155: Why have you added this check here? This is being handled in line 231.

xadupre: It fails in get_column_indices, this function maps names to integers and but it fails to find an input columns as they are dropped.

PR #270, opened 2019-09-25
Fixed output type for dtype=int64 in OHE

prabhat00155:


prabhat00155:
https://github.com/onnx/sklearn-onnx/issues/271

PR #262, opened 2019-09-19
[WIP] Add tests for lightgbm and xgboost in skl2onnx series

xadupre:

PR #276, opened 2019-09-30
[Proposal] Moves lightgbm and xgboost converters to skl2onnx

xadupre:
* easier to maintain
* both libraries follows scikit-learn API


prabhat00155:
I would prefer lightgbm and xgboost to be separate, mainly because of two reasons:
1) This is scikit-converter, so we should only have converters for scikit operators(https://github.com/scikit-learn/scikit-learn).
2) LightGBM fails to install on MacOS and this could cause issues to skl2onnx installation on MacOS. I already have to comment out `import lightgbm` code in one of the unit tests in order to run pytest.

xadupre:
Not sure why you get an error on lightgbm, the related tests are in a separate folder and there is no need to import it, that was the choice I made in the initial commit #262 already merged in master branch. If you look at the CI file, the first tests series are run before lightgbm and xgboost are installed to make sure it does not fail if they are missing.

As opposed to other models, SVM, sparkl, coremltools, we only support the scikit-learn API lightgbm of xgboost. That's why it makes sense to me to move the code in skl2onnx. Every change we make in the way we parse model has to be propagated to these converters and we did not do it in onnxmltools. That explains the issues users faced when they tried to insert lightgbm in a pipeline and convert it.

PR #281, opened 2019-10-07
Added BaggingClassifier converter

prabhat00155:


skl2onnx/operator_converters/bagging.py
```diff
@@ -0,0 +1,133 @@
...
+
+        container.add_initializer(
+            n_estimators_name, onnx_proto.TensorProto.FLOAT, [],
+            [len(model.estimators_)])
+        container.add_initializer(
+            class_labels_name, onnx_proto.TensorProto.INT64,
+            [1, 1, len(model.estimators_[0].classes_)],
+            model.estimators_[0].classes_)
+
+        container.add_node('Equal', [class_labels_name, merged_proba_name],
```
wenbingl: Should we add these two op into onnx_ops.apply_*, given they are popular?

prabhat00155: If there are multiple versions of the op, we may think of adding it.

PR #304, opened 2019-10-31
Flatten input data before using stringnormalizer for a better use in pipeline

xadupre:


tests/test_sklearn_tfidf_vectorizer_converter.py
```diff
@@ -29,7 +37,7 @@ def test_model_tfidf_vectorizer11(self):
         vect = TfidfVectorizer(ngram_range=(1, 1), norm=None)
         vect.fit(corpus.ravel())
         model_onnx = convert_sklearn(vect, "TfidfVectorizer",
-                                     [("input", StringTensorType([1]))],
+                                     [("input", StringTensorType())],
```
prabhat00155: Don't we need to pass shape?

skl2onnx/operator_converters/text_vectoriser.py
```diff
@@ -209,6 +209,15 @@ def convert_sklearn_text_vectorizer(scope, operator, container):
             "https://github.com/onnx/sklearn-onnx/issues.")
 
     if op.lowercase or op.stop_words_:
+
```
prabhat00155: What happens when the above if condition is False? flatten variable would be undefined, won't it? And that's being used below.

tests/test_sklearn_tfidf_vectorizer_converter.py
xadupre: Not always. My goal is to test batch predictions so I need to remove that information.

skl2onnx/operator_converters/text_vectoriser.py
xadupre: StringNormalizer is the only operator which does not accept 2D inputs (or inputs with more than one row). Flatten is not needed when this operator is not used.

prabhat00155: Check line 242, you are using flatten there(which would be undefined in the case I mentioned).

PR #308, opened 2019-11-05
Implement decision functions for classifiers

xadupre:


jiafatom:
Currently, the boolean raw_score is associated with classifier type, right?
If a model has two classifiers of same type (for example, SVM), is it possible one classifier uses raw score, the other classifier does not use raw score? Seems rare for this case though. But in this case, raw_score may be associated with the classifier instance, not the type.

xadupre:
Yes it is. Instead of settings options to ``{SVM: {'raw_scores': True}}``, you use ``{id(svm1): {'raw_scores': True}, id(svm2): {'raw_scores': True}}``.

PR #310, opened 2019-11-05
Sparse initializer

xadupre:
onnxruntime still does not support sparse.


skl2onnx/common/_container.py
```diff
@@ -267,26 +275,56 @@ def add_initializer(self, name, onnx_type, shape, content, can_cast=True):
...
+
         if isinstance(content, TensorProto):
             tensor = TensorProto()
             tensor.data_type = content.data_type
             tensor.name = name
             tensor.raw_data = content.raw_data
             tensor.dims.extend(content.dims)
+        elif (SparseTensorProto is not None and
+                isinstance(content, SparseTensorProto)):
+            raise NotImplementedError("Not implemented yet.")
```
prabhat00155: Should we also mention the link to file github issue like in other places?

xadupre: You're right.

Issue #321, opened 2019-11-13
Predictions seem wrong a scikit pipeline with OneHotEncoder


ivannp:
Have a simple model with two predictors:
* int64, plugged into a OneHot transform
* float32/float64, plugged into a RobustScaler

The model is created fine, but all predictions come out as 0s. None of them comes out as 0 in the original scikit model.

Removing the OneHot transform solves the issue - predictions are similar to the scikit model.

xadupre:
I'll close the issue as the PR is now merged into master branch.

victornoel:
@xadupre does the fix solves also the problem if the inputs of the `OneHotEncoder` are strings?

I'm facing a similar situation with one `OneHotEncoder` in front of a `XGBRegressor`: the output is of course not 0s (because of the regressor), but it is always the same value given.

victornoel:
@xadupre some clarification:
- I'm comparing the predictions via onnxruntime and sklearn/xgboost itself and the there are huge differences when using the `OneHotEncoder`
- I tried with the latest master of `sklearn-onnx` and the problem persists
- if I replace the `OneHotEncode` with an `OrdinalEncoder`, the predictions are equals between sklearn and onnxruntime

victornoel:
@xadupre have you been able to reproduce?

Just to be sure it wasn't related to some float32/float64 strange behaviour, I try to train my model using float32 and I still see huge differences.

For the record, the predictions from sklrean/xgb range from 0 to 1700 and those from onnxruntime range from 0 to 323, with diffs as high as 1400!

I'm not sure what is the best way to reproduce it in a small example, because I'm relying on proprietary data.

victornoel:
For the record here is the code I'm using for converting the model:
```py
train_df, test_df, train_labels, test_labels = train_test_split(
    full_df, full_labels, test_size=.2, random_state=None
)

col_transformer = _column_tranformer_fitted_from_df(full_df)

search = RandomizedSearchCV(
    XGBRegressor(verbose=0, objective='reg:squarederror'),
    param_distributions={
        "colsample_bytree": sp_uniform(),
        "gamma": sp_uniform(.1, 1),
        'learning_rate': sp_uniform(.1, .6),
        'max_depth': sp_randint(10, 30),
        'min_child_weight': sp_uniform(0, 3),
        'n_estimators': range(10, 75),  # nb de pas d'apprentissange pour chaque cas
    },
[...]

PR #322, opened 2019-11-14
Casts numerical input into int64 for OneHotEncoder (fixes #320, #321), changes parsing for ColumnTransformer

### LINKED ISSUE #320: OneHotEncoder works only with int64
Converting a scikit pipeline, I am able to use _OneHotEncoder_ only on a _int64_ column, but not on a _int32_. It should work on all column types supported by ONNX.

xadupre:
Onnx specifications is confusing about support of int32 as inputs. Attribute name for categories is cat_int64s and onnxruntime does not support int32 inputs.


ivannp:
Tried my failing test and it worked as expected.

xadupre:
Thanks for checking. The bug was not only with OneHotEncoder but also with ColumnTransformer, Pipeline as well.

Issue #312, opened 2019-11-07
Output Shape


XciD:
Hello,

I've followed the tutorial to Convert your scikit-learn model into ONNX,

If I understand well, the shape `[None, 4]` means that the first dimension is dynamic. In order to batch calls to the model.

The thing I don't understand is why the output tensor has a shape of `[1]` and not `[None]`

```python
input name='float_input' and shape=[None, 4]
NodeArg(name='output_label', type='tensor(int64)', shape=[1])
```

prabhat00155:
Can you give me link to the tutorial? I don't use output shape when converting models, so I'll need to look at the exact code.

XciD:
I've used this one, 

I've just replace: 
```
initial_type = [('float_input', FloatTensorType([1, 4]))]
``` 
to
```
initial_type = [('float_input', FloatTensorType([None, 4]))]
```

Seems like the first one don't work.
```
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Got invalid dimensions for input: float_input for the following indices
 index: 0 Got: 38 Expected: 1
 Please fix either the inputs or the model.
```

prabhat00155:
I don't see a link in your message, are you using this: http://onnx.ai/sklearn-onnx/?
Yes, we need to use _None_ for dynamic number of inputs because of changes in latest onnxruntime.

XciD:
Yes it’s this one. 
Yes I understood the None as input that create the dynamic shape. 

But the 1 as output should then be None ? As it’s a scalar output with a dynamic shape also ?

prabhat00155:
This is the correct example:

from sklearn.datasets import load_iris
from sklearn.model_selection import train_test_split
from sklearn.ensemble import RandomForestClassifier
iris = load_iris()
X, y = iris.data, iris.target
X_train, X_test, y_train, y_test = train_test_split(X, y)
clr = RandomForestClassifier()
clr.fit(X_train, y_train)

# Convert into ONNX format with onnxmltools
from skl2onnx import convert_sklearn
from skl2onnx.common.data_types import FloatTensorType
initial_type = [('float_input', FloatTensorType([None, 4]))]
onx = convert_sklearn(clr, initial_types=initial_type)
with open("rf_iris.onnx", "wb") as f:
    f.write(onx.SerializeToString())

# Compute the prediction with ONNX Runtime
import onnxruntime as rt
[...]

PR #323, opened 2019-11-15
Add a unit test to check onnxruntime output shape (investigate #312)

xadupre:


tests/test_shapes.py
```diff
@@ -0,0 +1,75 @@
...
+class TestSahpes(unittest.TestCase):
+
+    @unittest.skipIf(StrictVersion(onnx.__version__) < StrictVersion("1.6.0"),
+                     reason="not available")
+    @unittest.skipIf(StrictVersion(rt.__version__) < StrictVersion("1.0.0"),
+                     reason="not available")
+    def test_onnxruntime_shapes_reg(self):
+        iris = load_iris()
+        X, y = iris.data, iris.target
+        X_train, X_test, y_train, y_test = train_test_split(X, y)
```
prabhat00155: Can you use https://github.com/onnx/sklearn-onnx/blob/43e9c64380396fe71de7ab6dbb499cd52656134f/tests/test_utils/tests_helper.py#L51 for this? That would avoid code repetition.

tests/test_shapes.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @unittest.skipIf(StrictVersion(onnx.__version__) <= StrictVersion("1.6.0"),
+                     reason="not available")
+    @unittest.skipIf(StrictVersion(rt.__version__) <= StrictVersion("1.0.0"),
+                     reason="not available")
+    def test_onnxruntime_shapes_clr(self):
+        iris = load_iris()
+        X, y = iris.data, iris.target
+        X_train, X_test, y_train, y_test = train_test_split(X, y)
+        clr = RandomForestClassifier(max_depth=1)
```
prabhat00155: Same comment as above.

xadupre: I know but it is faster to analyse this way when it fails as you don't have to look into another file to understand what the data is.

Issue #326, opened 2019-11-26
Error in converting SageMaker model to onnx


csharpbd:
I'm trying to convert a SageMaker XGBoost model to ONNX. There is a good resource to use machine learning in business area. I've tried [Using Machine Learning to Improve Sales in SageMaker](http://fasttrackteam.com/using-machine-learning-to-improve-sales-a-simple-example.aspx) to create the model and then convert the model to ONNX model. The example is working well in SageMaker. I've load the model successfully by using `pickle` and `joblib`. But it's shown error, when I'm going to convert the model to onnx using `sklearn-onnx`. Please take a look at the attached images for more details about the error.

![sklearn-onnx #1](https://user-images.githubusercontent.com/25961704/69671898-89040880-105c-11ea-9e06-1bc0534a99ac.png)
![sklearn-onnx #2](https://user-images.githubusercontent.com/25961704/69671899-899c9f00-105c-11ea-911e-8dde494d656e.png)

Could you please take a look at the issues and let me know a way to solve the issues? Thanks in advance!

xadupre:
There is no solution yet. There is no converter in onnxmltools for a xgboost booster yet (see https://github.com/onnx/onnxmltools/issues/349).

csharpbd:
@xadupre thanks for your reply. I've tried the `winmltools` and `onnxmltools` but unfortunetly the tools are not working. Already I've posted a question on StackOverflow about it. 
You can find the question [Error in converting SageMaker XGBoost model to ONNX model
](https://stackoverflow.com/questions/59000779/error-in-converting-sagemaker-xgboost-model-to-onnx-model) I've tried to explain each steps that I followed to convert the mode to ONNX model. Could you please take a look at the question? I hope you may help me.

xadupre:
winmltools is using onnxmltools. That's why it does not work. I can try to have a fix by the end of the week. Would that work for you?

csharpbd:
Sure, that will help me a lot.
Thank you for your kindness.

xadupre:
I made a PR: https://github.com/onnx/onnxmltools/pull/355. You need to use onnxmltools from this branch: https://github.com/xadupre/onnxmltools/tree/xgbb. Let me know if it fixes your issue. I still have some doubts because it does not seem possible to retrieve the objective function from a booster. I only guess which kind of problem it is (classifier, regressor) but I might require an extra parameter if it does not work.

csharpbd:
[...]

PR #328, opened 2019-11-27
One step to #326, create plot_pipeline_xgboost.py

xadupre:
onnxmltools is still missing a converter for the xgboost booster.

### DIFF
--- a/docs/examples/plot_pipeline_xgboost.py
+++ b/docs/examples/plot_pipeline_xgboost.py
@@ -1,0 +1,156 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""
+.. _example-xgboost:
+
+Convert a pipeline with a XGBoost model
+========================================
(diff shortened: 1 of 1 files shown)


wenbingl:
Should this example be checked into ONNXMLTOOLS?

xadupre:
Maybe both? onnxmltools documentation is not really maintained.

docs/examples/plot_pipeline_xgboost.py
```diff
@@ -0,0 +1,156 @@
...
+your own pipeline, some of them are described
+and explained in :ref:`errors-pipeline`.
+
+.. contents::
+    :local:
+
+Train a XGBoost classifier
+++++++++++++++++++++++++++
+"""
+import xgboost
```
prabhat00155: Can you sort the import statments? It's not very readable this way, especially with so many of them. Can you group them into two: standard library imports followed by third party imports(and sort them within the group)?

PR #337, opened 2019-12-06
Pin sklearn version to fix broken onnxmltools tests

vinitra:
coremltools has not updated the breaking change to imputer in their libary for version 0.22 of sklearn.

As coremltools is an onnxmltools dependency, adding a max version pin for sklearn in requirements.txt temporarily.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,5 +3,5 @@ numpy>=1.15
 scipy>=1.0
 protobuf
 onnx>=1.2.1
-scikit-learn>=0.19
+scikit-learn>=0.19,<0.22
 onnxconverter-common>=1.5.1


xadupre:
sklearn-onnx works with 0.22, this issue should have been fixed on onxmltools repo.

vinitra:
onnxmltools has certain unit tests that use a coreml -> sklearn -> onnx pipeline. It also inherits its dependency on sklearn from skl2onnx. If the sklearn version updates, then the coremltools library (another dependency of onnxmltools) has errors. It seemed to make more sense to keep onnxmltools and sklearn-onnx consistent in terms of versioning and dependencies instead of adding a new sklearn requirement directly to onnxmltools separately from sklearn-onnx.

Hopefully this bug will be fixed shortly and then we can unpin the version.

PR #341, opened 2019-12-09
Proposal to convert HistGradientBoostingRegressor

xadupre:


skl2onnx/_supported_operators.py
```diff
@@ -168,8 +185,9 @@
     BernoulliNB, ComplementNB, GaussianNB, MultinomialNB,
     KNeighborsClassifier,
     CalibratedClassifierCV, OneVsRestClassifier, VotingClassifier,
-    AdaBoostClassifier, MLPClassifier, LinearDiscriminantAnalysis
-]
+    AdaBoostClassifier, MLPClassifier, LinearDiscriminantAnalysis,
+    HistGradientBoostingClassifier
```
prabhat00155: Comma at the end would help avoid modifying the last line when adding new elements.

skl2onnx/_supported_operators.py
```diff
@@ -216,8 +234,8 @@ def build_sklearn_operator_name_map():
                 GenericUnivariateSelect, RFE, RFECV, SelectFdr, SelectFpr,
                 SelectFromModel, SelectFwe, SelectKBest, SelectPercentile,
                 VarianceThreshold, GaussianMixture, GaussianProcessRegressor,
-                BayesianGaussianMixture, OneClassSVM,
-                PLSRegression
+                BayesianGaussianMixture, OneClassSVM, PLSRegression,
+                HistGradientBoostingClassifier, HistGradientBoostingRegressor
```
prabhat00155: Same comment as above.

skl2onnx/operator_converters/random_forest.py
```diff
@@ -25,7 +28,11 @@ def _num_estimators(op):
     #       ...
     #       classifier.fit(X_tmp, y_tmp)
     #       classifier.n_estimators += 30
-    return len(op.estimators_)
+    try:
+        return len(op.estimators_)
+    except AttributeError:
+        # HistGradientBoosting
+        return len(op._predictors)
```
prabhat00155: What if _predictors is not there in op? Let's write it in a way which makes extending it further easier without modification to existing code; something like this:
`if hasattr(op, 'estimators_'):`
`. num_est = len(op.estimators_)`
`elif hasattr(op, '_predictors'):`
`.  num_est = len(op._predictors)`
`else:`
`. # Either raise some error(NotImplementedError) if unexpected or set num_est to 0/None.`
`return num_est`

skl2onnx/operator_converters/random_forest.py
```diff
@@ -71,17 +78,31 @@ def convert_sklearn_random_forest_regressor_converter(scope,
     op_type = 'TreeEnsembleRegressor'
     attrs = get_default_tree_regressor_attribute_pairs()
     attrs['name'] = scope.get_unique_operator_name(op_type)
-    attrs['n_targets'] = int(op.n_outputs_)
+    try:
+        attrs['n_targets'] = int(op.n_outputs_)
+    except AttributeError:
+        # HistGradientBoostingRegressor
+        attrs['n_targets'] = op.n_trees_per_iteration_
```
prabhat00155: Again we can check for attributes using hasattr instead of try-except.

skl2onnx/operator_converters/random_forest.py
```diff
@@ -71,17 +78,31 @@ def convert_sklearn_random_forest_regressor_converter(scope,
...
-    tree_weight = 1. / estimtator_count
-    for tree_id in range(estimtator_count):
-        tree = op.estimators_[tree_id].tree_
-        add_tree_to_attribute_pairs(attrs, False, tree, tree_id,
-                                    tree_weight, 0, False, True,
-                                    dtype=container.dtype)
+    estimator_count = _num_estimators(op)
+    tree_weight = 1. / estimator_count
+    for tree_id in range(estimator_count):
+        if hasattr(op, 'estimators_'):
```
prabhat00155: Here you use hasattr, which seems the appropriate choice.

tests/test_sklearn_random_forest_converters.py
```diff
@@ -187,6 +196,41 @@ def test_model_multi_class_nocl(self):
                           "StrictVersion(onnxruntime.__version__)"
                           " <= StrictVersion('0.2.1')")
 
+    def common_test_model_hgb_regressor(self, add_nan=False):
```
prabhat00155: Where is the unit test for HistGradientBoostingClassifier converter?

xadupre: Below.

skl2onnx/_supported_operators.py
xadupre: We should short them all bug that's not the only place this is done like that.

tests/test_sklearn_random_forest_converters.py
xadupre: Not done yet in fact. In a different PR a in this one if this seems already ok.

PR #350, opened 2019-12-20
Add onnxruntime 1.1.0 to CI

xadupre:


tests_onnxmltools/test_lightgbm.py
```diff
@@ -16,6 +16,11 @@
 from skl2onnx.common.data_types import (
     SequenceType, DictionaryType, Int64TensorType, StringTensorType
 )
+import onnxmltools  # noqa
```
prabhat00155: Why these imports?

xadupre: To investigate which import of onnxmltools is failing.

prabhat00155: I see, you are not looking to check this in?

xadupre: I'd like to understand why onnxmltools is failing.

PR #369, opened 2020-01-15
Added decision function to one-vs-rest classifier converter

prabhat00155:


skl2onnx/operator_converters/one_vs_rest_classifier.py
```diff
@@ -99,9 +103,14 @@ def convert_one_vs_rest_classifier(scope, operator, container):
...
 
-        # normalizes the outputs
-        apply_normalization(scope, conc_name, operator.outputs[1].full_name,
-                            container, axis=1, p=1)
+        if use_raw_scores:
+            apply_identity(scope, conc_name,
+                           operator.outputs[1].full_name, container)
+        else:
+            # normalizes the outputs
+            apply_normalization(
```
xadupre: I wonder why we need normalisation here, the runtime should handle it as it is supposed to return probabilities.

prabhat00155: Removing normalisation causes incorrect probability scores.

PR #372, opened 2020-01-16
Shorten nightly build to be under 1h

xadupre:


.azure-pipelines/linux-CI-nightly.yml
```diff
@@ -114,12 +114,12 @@ jobs:
       pytest tests --durations=0 --basetemp=temp --doctest-modules --junitxml=junit/test-results.xml
     displayName: 'pytest'
 
-  - script: |
-      pip install onnxmltools openpyxl
-      coverage run --include=skl2onnx/** tests/benchmark.py
-      coverage report -m
-      coverage html
-    displayName: 'coverage'
+#  - script: |
```
prabhat00155: Why don't we just remove it instead?

xadupre: To avoid searching for the command line if we need to put it back.

prabhat00155: Don't we have the same statements in the CI build files?

wenbingl: I prefer to clean them up as well. you can check the git history if you want to see it again in the future.

xadupre: I'll do it but I doubt anybody will think of looking at the history to get this command line.

PR #373, opened 2020-01-20
Add a unit test to check a pipeline pca + rf

xadupre:


tests/test_sklearn_random_forest_converters.py
```diff
@@ -391,6 +396,29 @@ def test_model_extra_trees_classifier_multilabel_low_samples(self):
             "onnxruntime.__version__) <= StrictVersion('0.2.1')",
         )
 
+    @unittest.skipIf(not onnx_built_with_ml(),
+                     reason="Requires ONNX-ML extension.")
+    def test_boston_pca_rf(self):
+        data = load_boston()
```
prabhat00155: Why didn't you just use make_regression()?

xadupre: I did a demo and it failed with this specific example (with the latest released version). I wanted to check it was covered now.

PR #374, opened 2020-01-20
Add a test to check decision_function for LinearDiscriminantAnalysis

xadupre:


tests/test_sklearn_glm_classifier_converter.py
```diff
@@ -66,6 +66,52 @@ def test_model_logistic_linear_discriminant_analysis(self):
...
+            basename="SklearnLinearDiscriminantAnalysisBinRawScore-Out0",
+            # Operator cast-1 is not implemented in onnxruntime
+            allow_failure="StrictVersion(onnx.__version__)"
+                          " < StrictVersion('1.3') or "
+                          "StrictVersion(onnxruntime.__version__)"
+                          " <= StrictVersion('0.2.1')",
+            methods=['predict', 'decision_function']
+        )
+
+    @unittest.skipIf(not onnx_built_with_ml(),
```
prabhat00155: The two examples look exactly similar except one y value. Any reason for having them this way?

tests/test_sklearn_glm_classifier_converter.py
```diff
@@ -66,6 +66,52 @@ def test_model_logistic_linear_discriminant_analysis(self):
...
+        X_test = np.array([[-0.8, -1], [0, 1]], dtype=np.float32)
+        model = LinearDiscriminantAnalysis().fit(X, y)
+        model_onnx = convert_sklearn(
+            model, "linear model",
+            [("input", FloatTensorType([None, X_test.shape[1]]))],
+            options={id(model): {'raw_scores': True}})
+        self.assertIsNotNone(model_onnx)
+        dump_data_and_model(
+            X_test, model, model_onnx,
+            basename="SklearnLinearDiscriminantAnalysisBinRawScore-Out0",
```
prabhat00155: Do we need Out0, when we have specified the methods?

tests/test_sklearn_glm_classifier_converter.py
xadupre: Just to check binary and multiclass are ok.

tests/test_sklearn_glm_classifier_converter.py
xadupre: Shape mismatch otherwise.

PR #377, opened 2020-01-22
Fix #301, support logistic regression in CalibratedClassifierCV

### LINKED ISSUE #301: CalibratedClassifierCV returns wrong predictions when used in pipeline
When using the CalibratedClassifierCV in a Pipeline that is exported to ONNX, then running the ONNX session returns completely different results than running the pipeline directly.

To reproduce the issue, take the tutorial example "Convert a pipeline with ColumnTransformer" (http://onnx.ai/sklearn-onnx/auto_examples/plot_complex_pipeline.html) and replace

clf = Pipeline(steps=[('preprocessor', preprocessor),
                      ('classifier', LogisticRegression(solver='lbfgs'))])

with 

clf = Pipeline(steps=[('preprocessor', preprocessor),
                      ('calibrator', CalibratedClassifierCV(base_estimator=LogisticRegression(solver='lbfgs', random_state=42), method='sigmoid', cv='warn'))])  

Then predict the same row once using the classifier and once using the ONNX model. The predicted probabilities are completely different.

(Attached is also the complete code).
[titanic.zip](https://github.com/onnx/sklearn-onnx/files/3784470/titanic.zip)

xadupre:


skl2onnx/_supported_operators.py
```diff
@@ -193,11 +193,6 @@
 # each cluster in most cases.
 cluster_list = [KMeans, MiniBatchKMeans]
 
-# Classifiers with converters supporting decision_function().
```
prabhat00155: I don't think we should remove this unless we support all the classifiers with decision_function(). I am working on it though.

xadupre: I think we should test whether or not an option is available but not add a specific mechanism for classifiers supporting decision_function. That's the first Reason I removed it, the second reason is we forgot to update it. So it was wrong. I really prefer to address the fact we check option. It is more generic.

prabhat00155: CalibratedClassifierCV doesn't support decision_function() itself, so it's not about checking options that user passes. It should always call decision_function() if the base_classifier has decision_function(), else it should call predict_proba(). I added this check since, we didn't support decision_function for most. Now that we have that, we should add a check if the base_classifier has decision_function attribute and if it does call them accordingly.

prabhat00155: Since, you are doing this:
container.add_options(id(base_model), {'raw_scores': True})
we don't need to check for the presence of decision_function(), as we only use this(check raw_scores) in classifiers converters that have decision_function. We may want to note that here though(in comments).

xadupre: See PR #380.

skl2onnx/operator_converters/calibrated_classifier_cv.py
```diff
@@ -237,6 +236,10 @@ def convert_calibrated_classifier_base_estimator(scope, operator, container,
                 operator.raw_operator.__class__.__name__))
 
     base_model = model.base_estimator
+    if not hasattr(base_model, "predict_proba"):
```
prabhat00155: Why this check? Also you are checking for predict_proba and error says no decision_function.

xadupre: Fixed

PR #381, opened 2020-01-22
Stacking classifier converter

prabhat00155:


prabhat00155:
Checked in as a part of https://github.com/onnx/sklearn-onnx/pull/382.

PR #397, opened 2020-02-04
Improves robustness of knn

xadupre:


skl2onnx/operator_converters/nearest_neighbours.py
```diff
@@ -79,13 +82,19 @@ def onnx_nearest_neighbors_indices(X, Y, k, metric='euclidean', dtype=None,
             [-1], dtype=dtype), op_version=op_version)
         node = OnnxTopK_10(neg_dist, np.array([k], dtype=np.int64),
                            op_version=10, **kwargs)
-    else:
+    elif largest0:
```
prabhat00155: Why do we need this argument(largest0)? This is only available in opset 11 right, we can always use that for opset 11.

xadupre: True, I did it that way to check it was ok. The goal is to remove it. There is still an issue with neighbours at the exact same distance.

prabhat00155: Yeah, not sure if we can do much about the issue with equidistant neighbours(it may be acceptable as well, similar to classes with equal probability scores). Do you want to remove largest0? Can you check if passing largest=0 always works for topk opset 11(if yes, that should be sufficient).

Issue #376, opened 2020-01-21
support for PowerTransformer


msteijaert:
I noticed that there is not yet support for PowerTransformer. Is anybody working on this?

If not, I will see if I can solve this on my end. I'm new to this package, but here is my naive understanding of what needs to happen (feel free to correct or comment):

1) creation of a scaler (at first glance the new converter can be almost identical to calculate_sklearn_function_transformer_output_shapes or calculate_sklearn_scaler_output_shapes)

2) creation of a converter, which reimplements PowerTransformer.transform(), and makes use of the method name and fitted lambda values stored in operator object

xadupre:
You can look into PR #318 or PR #294 depending on what you need to do. The first step consists in compute the shape of the output given the input. You will find more explanation in this example: http://onnx.ai/sklearn-onnx/auto_examples/plot_custom_model.html.

xadupre:
I'll close this issue as the PR was merged.

PR #403, opened 2020-02-12
Adding converter for PowerTransformer

Enolerobotti:
Regarding the issue #376


skl2onnx/operator_converters/__init__.py
```diff
@@ -39,6 +39,7 @@
 from . import one_vs_rest_classifier
 from . import ordinal_encoder
 from . import polynomial_features
+from . import powertransformer
```
xadupre: I would choose power_transformer.py to follow what was previously done.

skl2onnx/operator_converters/powertransformer.py
```diff
@@ -0,0 +1,131 @@
...
+
+def convert_powertransformer(scope, operator, container):
+    """Converter for PowerTransformer"""
+    op_in = operator.inputs[0]
+    op_out = operator.outputs[0].full_name
+    op = operator.raw_operator
+    lambdas = op.lambdas_
+
+    # tensors of units and zeros
+    ones_ = OnnxDiv(op_in, op_in)
```
xadupre: You need to specify op_version=opv where opv = container.target_opset (see https://github.com/onnx/sklearn-onnx/blob/master/skl2onnx/operator_converters/k_means.py#L79). Otherwise, the converter only picks the latest version of every operator.

Enolerobotti: Ok, Thank you!

PR #404, opened 2020-02-12
Added KNNImputer converter

prabhat00155:


tests/test_sklearn_nearest_neighbour_converter.py
```diff
@@ -489,6 +492,32 @@ def test_sklearn_k_neighbours_transformer_connectivity(self):
...
+                     reason="ConstantOfShape op not available")
+    def test_sklearn_knn_imputer(self):
+        x_train = numpy.array(
+            [[1, 2, numpy.nan, 12], [3, numpy.nan, 3, 13],
+             [1, 4, numpy.nan, 1], [numpy.nan, 4, 3, 12]], dtype=numpy.float32)
+        x_test = numpy.array(
+            [[1.3, 2.4, numpy.nan, 1], [-1.3, numpy.nan, 3.1, numpy.nan]],
+            dtype=numpy.float32)
+        model = KNNImputer(n_neighbors=3, metric='nan_euclidean').fit(x_train)
+        model_onnx = convert_sklearn(
```
xadupre: You should test against opset 9, 10, 11 (see unittest.subTest(opset=opset) to split a unit test wirthout splitting the function).

prabhat00155: You mean call convert_sklearn(), passing opset=9,10 and 11?

xadupre: yes, I know it is being somehow tested in different CI jobs but it can also be tested in a single job and make sure we generate a viable ONNX with the last onnx package if target_opset is 9, 10, 11.

prabhat00155: Added.

Issue #402, opened 2020-02-12
Support drop option of OneHotEncoder


victornoel:
Hi,

I tried to use a `OneHotEncoder` with `drop` option set to `first` but when I do so, running the generated model with onnxruntime gives me  the following error during prediction:
```
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running OneHotEncoder node. Name:'OneHotEncoder7' Status Message: Unknown Category and zeros = 0.
```

See https://github.com/onnx/sklearn-onnx/issues/321#issuecomment-584052819 for an example of code used to generate and convert the model.

prabhat00155:
I think you are passing an unknown category(category in test set not seen in training). Have you checked scikit output, that should give you an error(related to unknown category) as well.

victornoel:
@prabhat00155 I fit the `OneHotEncoder` with the full train+test dataset to avoid such problem. With sklearn it works as expected (`full_df` in https://github.com/onnx/sklearn-onnx/issues/321#issuecomment-584052819)

victornoel:
@prabhat00155 tomorrow I will try to make a repro using one of the sklearn-provided datasets :)

victornoel:
@prabhat00155 here it is: https://gist.github.com/victornoel/c4da3089f6792ac1b6fdcdb583886f97

prabhat00155:
To simplify, I extracted the text columns separately and just did OHE transform.
```
data = pd.read_csv(('https://raw.githubusercontent.com/amueller/'
                        'scipy-2017-sklearn/091d371/notebooks/datasets/titanic3.csv'))
text_df3 = data[data.columns[data.dtypes == 'object']]
list(map(lambda col: text_df3[col].fillna('N/A', inplace=True), text_df3.columns))
print(text_df3.isnull().values.any())
model = OneHotEncoder().fit(text_df3)
onnx_model = convert_sklearn(model, 'ohe', [('input', StringTensorType([None, text_df3.values.shape[1]]))])
save_model(onnx_model, 'ohe.onnx')
sess = InferenceSession('ohe.onnx')
res = sess.run(None, input_feed={'input': text_df3.values})
res
```
This works on the whole dataset, but if I divide the dataset into train and test:
```
train_df, test_df = train_test_split(text_df3, test_size=.2)
model = OneHotEncoder().fit(train_df)
model.transform(test_df)

[...]

PR #405, opened 2020-02-18
Fixed OHE bug when drop is set

prabhat00155:


prabhat00155:
Fixes https://github.com/onnx/sklearn-onnx/issues/402.

PR #419, opened 2020-03-17
Implemented SVM converters using primitive ops

prabhat00155:


prabhat00155:
Still need to implement converters for OneClassSVM and SVC models.

Issue #343, opened 2019-12-10
Error with multi-label classification and zipmap set to False


prabhat00155:
MLPClassifier and OneVsRestClassifier converters throw error with multi-label classification when zipmap is set to False.
```
X, y = make_multilabel_classification(n_labels=5, n_classes=10)
X = X.astype(np.float32)
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.5, random_state=42)
model = MLPClassifier().fit(X_train, y_train)
options = {id(model): {'zipmap': False}}
model_onnx = convert_sklearn(model, 'mlp', [('input', FloatTensorType([None, X_test.shape[1]]))], options=options)
save_model(model_onnx, 'mlp.onnx')
sess = InferenceSession('mlp.onnx')
res = sess.run(None, input_feed={'input': X_test})
```

Fail                                      Traceback (most recent call last)
<ipython-input-22-501a013b9cdb> in <module>
      7 model_onnx = convert_sklearn(model, 'mlp', [('input', FloatTensorType([None, X_test.shape[1]]))], options=options)
      8 save_model(model_onnx, 'mlp.onnx')
----> 9 sess = InferenceSession('mlp.onnx')
     10 res = sess.run(None, input_feed={'input': X_test})
     11 print(np.mean(np.isclose(model.predict_proba(X_test), list(map(lambda x:

~/Documents/MachineLearning/onnx_projects/tmp_env/lib/python3.6/site-packages/onnxruntime/capi/session.py in __init__(self, path_or_bytes, sess_options)
     21         self._path_or_bytes = path_or_bytes
     22         self._sess_options = sess_options
---> 23         self._load_model()
     24         self._enable_fallback = True
     25 

~/Documents/MachineLearning/onnx_projects/tmp_env/lib/python3.6/site-packages/onnxruntime/capi/session.py in _load_model(self, providers)
     33 
     34         if isinstance(self._path_or_bytes, str):
---> 35             self._sess.load_model(self._path_or_bytes, providers)
     36         elif isinstance(self._path_or_bytes, bytes):
     37             self._sess.read_bytes(self._path_or_bytes, providers)

Fail: [ONNXRuntimeError] : 1 : FAIL : Load model from mlp.onnx failed:Type Error: Type (tensor(int64)) of output arg (label) of node (N8) does not match expected type (tensor(float)).

xadupre:
It seems to be working now (see #443). Maybe it was fixed by another commit.

[...]

PR #443, opened 2020-04-28
Moves one unused initialializer in MLP converter, addresses #343

xadupre:


skl2onnx/operator_converters/multilayer_perceptron.py
```diff
@@ -110,9 +110,6 @@ def convert_sklearn_mlp_classifier(scope, operator, container):
     else:
         classes = np.array([s.encode('utf-8') for s in classes])
 
-    container.add_initializer(classes_name, class_type,
```
prabhat00155, line 113: You may also want to move the declaration on line 98.

xadupre: done

PR #450, opened 2020-04-29
Fixes #i302, handle OVR decision function

xadupre:


skl2onnx/operator_converters/support_vector_machines.py
```diff
@@ -140,6 +148,127 @@ def convert_sklearn_svm(scope, operator, container):
         raise ValueError("Unknown support vector machine model type found "
                          "'{0}'.".format(operator.type))
 
+    if (hasattr(op, 'decision_function_shape') and
+            op.decision_function_shape == 'ovr' and handles_ovr):
+        # Applies _ovr_decision_function.
```
prabhat00155, line 153: Wouldn't it be better to just add reference to the scikit implementation rather than adding the entire code here?

skl2onnx/operator_converters/support_vector_machines.py
```diff
@@ -140,6 +148,127 @@ def convert_sklearn_svm(scope, operator, container):
...
+        #             k += 1
+        #     transformed_confidences = (
+        #         sum_of_confidences / (3 * (np.abs(sum_of_confidences) + 1)))
+        #     return votes + transformed_confidences
+        if list(op.classes_) != list(range(len(op.classes_))):
+            raise RuntimeError(
+                "Classes different from first n integers are not supported "
+                "in SVC converter.")
+
+        cst3 = scope.get_unique_variable_name('cst3')
```
prabhat00155, line 181: It's difficult to understand what cst0, cst1 and cst3 mean. It would be better to have clearer node names.

xadupre: cst0 = constant equal to 0, cst1 = constant equal to 1, cst3 = constant equal to 3

skl2onnx/operator_converters/support_vector_machines.py
xadupre: links to code are not really stable and it is easier to understand what the converter does by looking at the numpy code

skl2onnx/operator_converters/support_vector_machines.py
prabhat00155: Variable name is fine, it would be better to have the onnx node name as `const0, etc.` Anyways, this is a minor issue.

PR #454, opened 2020-05-04
Fixes #451, enables white list, back list for operator types

### LINKED ISSUE #451: Possibility to Customize Conversion Based on Operator
Hi, as you already add scoring method in GMM model for sklearn. The scoring method worked well when I used OnnxRuntime.

Now I got a new problem when I implemented my model to the new platform in this case Unity, where Unity could only load ONNX using Baraccuda with a limited type of operator.

Baraccuda in Unity doesn't contain operator of ReduceLogSumExp and ReduceSumSquare that appear in my model. For now, Baraccuda then suggest me to change this operator in onnx model to:

x -> y = exp(x) -> z = reduce_sum(y) -> log(z) for ReduceLogSumExp
and
x -> y = x^2 -> z = reduce_sum(y) for ReduceSumSquare

Could I customize the conversion so ReduceLogSumExp automatically just only used Exp and reduce_sum and reducesumsquare used square and reduce sum?

Thank You for your kind support.

And here I attached the visualization of my onnx model:
image

xadupre:


skl2onnx/common/_container.py
```diff
@@ -95,17 +95,52 @@ def _build_options(model, defined_options, default_values, allowed_options):
...
+            return True
+        except RuntimeError:
+            return False
+
+    def check_white_black_list(self, node_type):
+        """
+        Checks a node type is allowed according to white
+        and black lists.
+        """
+        if self._white_op:
```
prabhat00155, line 121: You could merge the two ifs into one.

skl2onnx/common/_container.py
```diff
@@ -95,17 +95,52 @@ def _build_options(model, defined_options, default_values, allowed_options):
...
+            if node_type not in self._white_op:
+                raise RuntimeError(
+                    "Operator '{}' is not white listed.".format(node_type))
+        if self._black_op:
```
prabhat00155, line 125: Same as above.

skl2onnx/operator_converters/gaussian_mixture.py
```diff
@@ -34,6 +36,8 @@ def convert_sklearn_gaussian_mixture(scope, operator, container):
     opv = container.target_opset
     options = container.get_options(op, dict(score_samples=None))
     add_score = options.get('score_samples', False)
+    combined_reducesum = not container.is_allowed(
+        {'ReduceLogSumExp', 'ReduceSumSquare'})
```
prabhat00155, line 40: Why are we doing this?

skl2onnx/common/_container.py
```diff
@@ -177,25 +214,32 @@ def add_output(self, variable):
...
+                 white_op=None, black_op=None):
         """
         :param target_opset: number, for example, 7 for *ONNX 1.2*, and
                              8 for *ONNX 1.3*.
         :param dtype: float type to be used for every float coefficient
         :param options: see :ref:`l-conv-options`
         :param registered_models: registered models
+        :param white_op: white list of ONNX nodes allowed
+            while converting a pipeline, if empty, all are allowed
+        :param black_op: black list of ONNX nodes allowed
```
prabhat00155, line 235: Can you help me understand white_op and black_op? Do you exclude black_op nodes from model conversion?

xadupre: Some runtime do not implement all operators, a converter may know the list of available operators. The conversion raises an exception if one converter uses an operator not allowed. Actually, only the converter for GaussianMixture changes its behaviour not to use ReduceLogSumExp.

skl2onnx/common/_container.py
```diff
@@ -95,17 +95,52 @@ def _build_options(model, defined_options, default_values, allowed_options):
...
     It provides an abstract layer so that our parsing
     framework can work with models generated by different tools.
     """
 
-    def __init__(self, raw_model, dtype):
+    def __init__(self, raw_model, dtype, white_op=None, black_op=None):
         """
         :param raw_model: *scikit-learn* model to convert
         """
+        _WhiteBlackContainer.__init__(
```
wenbingl, line 142: not super?

skl2onnx/common/_container.py
xadupre: I thought that it would be better for the user to know which list is involved in the error message.

skl2onnx/operator_converters/gaussian_mixture.py
xadupre: Issue raised by a user.

skl2onnx/common/_container.py
xadupre: Same answer.

PR #463, opened 2020-05-14
Fixes #448, test raw_scores for OneVsRestClassifier

### LINKED ISSUE #448: OneVsRestClassifier prediction mismatch when raw_scores=True
```
import numpy as np
from sklearn.datasets import make_classification
from sklearn.model_selection import train_test_split
from sklearn.multiclass import OneVsRestClassifier
from onnxmltools import save_model
from onnxruntime import InferenceSession
from skl2onnx.convert import convert_sklearn

X, y = make_classification(n_classes=2, n_samples=1000, random_state=42, n_features=100, n_informative=7)

X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.5, random_state=42)
model = OneVsRestClassifier(estimator=GradientBoostingClassifier(random_state=42)).fit(X_train, y_train)

options = {id(model): {'raw_scores': True}}
onnx_model = convert_sklearn(model, 'lr', [('input', FloatTensorType([None, X_test.shape[1]]))], options=options)
save_model(onnx_model, 'ovr.onnx')
sess = InferenceSession('ovr.onnx')
res = sess.run(None, input_feed={'input': X_test.astype(np.float32)})
print(np.mean(res[0] == model.predict(X_test)))
```
>> 0.514

ONNX model always predicts 1 here as it does an argmax on decision function scores, not probabilities(when raw_scores=True).

xadupre:


tests/test_sklearn_one_vs_rest_classifier_converter.py
```diff
@@ -388,8 +391,34 @@ def test_ovr_regression_int_ensemble(self):
             model_onnx,
             basename="SklearnOVRRegressionIntEnsemble-Out0",
             allow_failure="StrictVersion(onnxruntime.__version__)"
-            "<= StrictVersion('0.2.1')",
-        )
+            "<= StrictVersion('0.2.1')")
```
prabhat00155, line 394: We already have tests for OVR with raw_scores=True, why add this?

tests/test_sklearn_one_vs_rest_classifier_converter.py
```diff
@@ -388,8 +391,34 @@ def test_ovr_regression_int_ensemble(self):
...
+
+    @unittest.skipIf(not onnx_built_with_ml(),
+                     reason="Requires ONNX-ML extension.")
+    @unittest.skipIf(StrictVersion(ort_version) < StrictVersion("1.2.0"),
+                     reason="fails to load the model")
+    def test_ovr_raw_scores(self):
+        X, y = make_classification(
```
prabhat00155, line 401: We coud use fit_classification_model() and avoid code-repetition.

xadupre: Code repetition is not always bad. In that case, the reader knows exactly what the test does. In my case, fit_classification_model does not expose all the parameters I need. I could expose them or keep it that way.

prabhat00155: You are using n_samples=100, rest seem to be the same. You may as well use 1000 used in fit_classification_model() or add an additional argument. I don't see why you won't be able to use fit_classification_model() here, unless I am missing something obvious.

PR #464, opened 2020-05-19
Fixes #462, fix issues in converter ArrayFeatureExtractor, OneHotEncoder

### LINKED ISSUE #462: Possible bug in convert_sklearn_array_feature_extractor(), operator.inputs not reshaped
Hi, 
before the issue report: skl2onnx is just great, so much better than pickling sklearn models.

/Users/XXXX/.conda/envs/cfserver/lib/python3.7/site-packages/sklearn/mixture/gaussian_mixture.py:366: RuntimeWarning: divide by zero encountered in log n_components, -1)[:, ::n_features + 1]), 1)) 
came up when I converted a pipeline. The resulting onnx model does not work properly and returns '0' as prediction results. I think it is caused when converting a ColumnTransformer in a 
pipeline because convert_sklearn_array_feature_extractor() does not reshape the inputs variable size as in other converters. 

I'm by no means an onnx expert but tried to dig through the code and attached an explanation of what I think happens and a code snippet that in essence replicates the issue.

Cheers

[column_transformer_bug.txt](https://github.com/onnx/sklearn-onnx/files/4626917/column_transformer_bug.txt)
[column_transformer_snippet_py.txt](https://github.com/onnx/sklearn-onnx/files/4626918/column_transformer_snippet_py.txt)

xadupre:


skl2onnx/operator_converters/one_hot_encoder.py
```diff
@@ -90,9 +90,28 @@ def convert_sklearn_one_hot_encoder(scope, operator, container):
...
+                    raise RuntimeError(
+                        "Category '{}' cannot be casted into int.".format(c))
+                if ci != c:
+                    raise RuntimeError(
+                        "Category '{}' is not an int64.".format(c))
+            attrs['cats_int64s'] = categories.astype(np.int64)
+        else:
+            raise RuntimeError(
+                "Input type {} is not supported for OneHotEncoder. "
+                "Ideally, it should either integer or strings.".format(
```
prabhat00155, line 113: You are missing `be`(it should either `be` integer or string).

PR #467, opened 2020-05-20
Fixes #466, support of zipmap option in BaggingClassifier

### LINKED ISSUE #466: BaggingClassifier does not support option zipmap: False

xadupre:


xadupre:
Already merged.

PR #473, opened 2020-06-02
More consistent opset accross converters and unit tests

xadupre:


xadupre:
Adresses #474.

skl2onnx/common/_container.py
```diff
@@ -60,7 +60,8 @@ def _get_operation_list():
                     # does not follow the same schema
                     continue
                 if k in {'apply_less_or_equal', 'apply_greater_or_equal',
-                         'apply_relu6'}:
+                         'apply_relu6', 'apply_relu_6', 'apply_slice2',
```
wenbingl, line 63: so every changes on apply_functions will cause an update here?

xadupre: I added this test when these functions were part of skl2onnx to do basic verification on the function code. I should remove it. There is no unit test in onnxconverter-common so this was added to catch some common mistakes.

PR #475, opened 2020-06-07
Implements pickable options, propagate common options in a pipeline

xadupre:


skl2onnx/_parse.py
```diff
@@ -450,6 +452,127 @@ def parse_sklearn(scope, model, inputs, custom_parsers=None, final_types=None):
     return outputs
 
 
+def parse_sklearn(scope, model, inputs, custom_parsers=None, final_types=None):
+    """
+    This is a delegate function. It does nothing but invokes the
+    correct parsing function according to the input model's type.
```
wenbingl, line 458: Both of them are the delegate function? I saw this comment twice?

skl2onnx/common/onnx_optimisation_identity.py
```diff
@@ -0,0 +1,116 @@
+"""
+Optimization of :epkg:`ONNX` graphs.
+Functions in *onnxconverter-common* do not support
```
wenbingl, line 3: onnxconverter-common does support opset.version < 9 with a limited optimization capability. But I think identity clean up should be included.

xadupre: It failed when I tried to use it.

skl2onnx/_parse.py
xadupre: Very old comment. I need to check.

skl2onnx/common/onnx_optimisation_identity.py
wenbingl: Can you the pre-optimized ONNX model?  we can take a look and fix it if any.

xadupre: Ok. I'll switch to onnxconverter_common optimizer and go through all unit tests.

PR #497, opened 2020-06-26
Fixes #496, merges exact same initializers

### LINKED ISSUE #496: Removes duplicated constant in container

xadupre:


skl2onnx/algebra/type_helper.py
```diff
@@ -31,6 +31,9 @@ def _guess_dim(value):
 
     if isinstance(given_type, (np.ndarray, coo_matrix)):
         shape = list(given_type.shape)
+        if len(shape) == 0:
```
guschmue, line 34: if not shape: ?

xadupre: I usually avoid not shape or shape as I know shape is not None. That means testing one less condition.

guschmue: pylint would complain about it :)

xadupre: flake8 is used instead of pylint. We can switch but maybe in another PR as I expect many lint issues to pop up.

guschmue: no worry, I'm nit picking.

PR #551, opened 2020-10-29
Fix library and examples when used with onnx 1.8.0

xadupre:


docs/examples/plot_intermediate_outputs.py
```diff
@@ -40,10 +40,10 @@
 from skl2onnx.helpers.onnx_helper import load_onnx_model
 import numpy
 import onnxruntime as rt
-from skl2onnx import convert_sklearn
+from skl2onnx import convert_sklearn, __max_supported_opset__ as max_opset
```
wenbingl, line 43: this name looks very internal, is it a good idea to be public?

xadupre: With this PR, sklearn-onnx is tested with onnx 1.7 (opset 12) or onnx 1.8 but up to opset 12. I can replace this name by a constant, 12 to avoid using it.

PR #565, opened 2020-12-03
feat: Add support for boolean classes

mjamei:


skl2onnx/_parse.py
```diff
@@ -368,6 +368,8 @@ def _parse_sklearn_classifier(scope, model, inputs, custom_parsers=None):
         this_operator.classlabels_int64s = classes
     elif np.issubdtype(classes.dtype, np.unsignedinteger):
         this_operator.classlabels_int64s = classes
+    elif np.issubdtype(classes.dtype, np.bool):
```
xadupre, line 371: A unit test should be added to cover that case.

xadupre:
This is no longer an issue with the latest version. Closing the PR.

PR #632, opened 2021-03-22
relicense to Apache License v2

szha:
Signed-off-by: Sheng Zha <zhasheng@amazon.com>

### DIFF
--- a/skl2onnx/algebra/automation.py
+++ b/skl2onnx/algebra/automation.py
@@ -1,8 +1,5 @@
-# -------------------------------------------------------------------------
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License. See License.txt in the project root for
-# license information.
-# --------------------------------------------------------------------------
+# SPDX-License-Identifier: Apache-2.0
+
 import textwrap
 import onnx
 import onnx.defs  # noqa
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,202 @@
-MIT License
 
-Copyright (c) 2018 Microsoft Corporation
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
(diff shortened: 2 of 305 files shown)


prasanthpul:
This is part of the ONNX-wide relicensing that has already been done by the other converter repos.

skl2onnx/algebra/automation.py
```diff
@@ -1,8 +1,5 @@
-# -------------------------------------------------------------------------
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License. See License.txt in the project root for
-# license information.
-# --------------------------------------------------------------------------
+# SPDX-License-Identifier: Apache-2.0
```
prasanthpul, line 1: @szha I noticed the copyrights in many files were removed as part of the license change. Was it on purpose or unintentionial?

szha: This is intentional. We can include the copyright notices into a unified NOTICE file for simplicity.

xadupre (approved):

faxu:
@szha @xadupre  this also needs to be updated in the documentation on this page: http://onnx.ai/sklearn-onnx/

szha:
@faxu where is the source of this website?

faxu:
I believe it's here: https://github.com/onnx/sklearn-onnx/blob/gh-pages/index.html

But it should be generated using the .rst file [here](https://github.com/onnx/sklearn-onnx/blob/master/docs/index.rst) which already reflects the license update. 

@xadupre can you help to regenerate the html pages based on the latest updates?

xadupre:
Source of the website are in branch gh-pages. Documentation is generated using sphinx (same as onnxruntime). You can use the command line ``python -m sphinx -j1 -v -N -T -b html -d build/doctrees docs dist/html``.

PR #664, opened 2021-06-10
Enables opset 14 and enables onnxruntime 1.8.0 on CI, remove CI for onnxruntime<1.5.0

xadupre:
onnxruntime implements supports opset 14, with this PR, CI tests all converters against opset 14.

### DIFF
--- a/.azure-pipelines/linux-conda-CI.yml
+++ b/.azure-pipelines/linux-conda-CI.yml
@@ -255,6 +255,8 @@ jobs:
         export TEST_TARGET_OPSET="$(onnx.target_opset)"
       fi
       cd tests
+      python -c "from onnxruntime import __version__;print('onnxruntime', __version__)"
+      python -c "from onnx import __version__;print('onnx', __version__)"
       python -c "from skl2onnx import __max_supported_opset__;print('__max_supported_opset__', __max_supported_opset__)"
       python -c "import test_utils;print('max_onnxruntime_opset', test_utils.max_onnxruntime_opset())"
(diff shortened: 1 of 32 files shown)


jiafatom:
> Opset 14 is not tested as onnxruntime does not implement this opset yet.

Change the description?

jiafatom:
There are several build errors seems related:
```python
           onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Failed to load model with error: /onnxruntime_src/onnxruntime/core/graph/model_load_utils.h:47 void onnxruntime::model_load_utils::ValidateOpsetForDomain(const std::unordered_map<std::basic_string<char>, int>&, const onnxruntime::logging::Logger&, bool, const string&, int) ONNX Runtime only *guarantees* support for models stamped with official released onnx opset versions. Opset 14 is under development and support for this is limited. The operator schemas and or other functionality may change before next ONNX release and in this case ONNX Runtime will not guarantee backward compatibility. Current official support for domain ai.onnx is till opset 13.
.```

xadupre:
> There are several build errors seems related:

I need to change the target_opset in many unit tests. By default, it takes onnx opset, 14, but if no used operators were updated in 14, the final opset was lower. Since Add, Sub, Mul are used in many opset, the final opset is usually 14. So the combination onnx==1.9.0 + onnxruntime==1.7.0 makes many unit tests fail. I already did many of them a couple of weeks ago. This is the second wave.

jiafatom (approved):

Issue #670, opened 2021-06-26
please include type hint at least in the tutorial/documentation


busFred:
To whom it may concern,

I implemented an algorithm following the sklean API. However, following implementing the onnx transformer is extremely difficult due to the confusion with the datatype.
In this [example](https://onnx.ai/sklearn-onnx/auto_tutorial/plot_icustom_converter.html), what is the datatype of input argument operator. Without the type information, what instance variables does the "operator" have and what method are available to us is extremely difficult to investigate. In other words, the difficulty of developing and making contribution is increased significantly.

I totally understand updating the entire coding project and annotate it with type hint can be difficult, but is it possible just include some of the info in the documentation?

Sincerely,

Hung-Tien Huang

xadupre:
I'll try to do that. Something similar to what you did [here](https://github.com/busFred/sklearn_plugins/blob/master/src/sklearn_plugins/cluster/_onnx_transform.py#L50) would work?

busFred:
> 
> 
> I'll try to do that. Something similar to what you did [here](https://github.com/busFred/sklearn_plugins/blob/master/src/sklearn_plugins/cluster/_onnx_transform.py#L50) would work?

Yes, exactly. Appreciate it

xadupre:
Closing the issue. The PR was merged and a new release was published.

PR #677, opened 2021-07-01
Add annotation to converters and shape calculator

xadupre:
Addresses #670.


Ark-kun:
This PR seems to make 665 changes. Does this PR only add annotation to converters and shape calculator?

xadupre:
Yes it is only about that.

PR #697, opened 2021-08-06
Support certain special raw names

janjagusch:
We have a use case where we'd like to support certain special characters in the `variable.raw_name` (in our case: `(`, and `-`).

I noticed that there seems to be a small bug in the `convert_topology` function, where inputs are only added to the container if their `raw_name` is equal to their `onnx_name`. I believe I fixed that.

Also similar to https://github.com/onnx/sklearn-onnx/pull/687, I removed the check that raises an error when the `raw_name` contains a `(`. Again, we have a use case where being able to have `(` as part of the raw name would be extremely useful and I don't see a reason why this would break the ONNX graph. If this is an oversight on my end, please let me know. 🙂

I also added some test cases to make sure everything is working as intended.

If this contribution gets accepted, I would suggest to close https://github.com/onnx/sklearn-onnx/pull/696.

### DIFF
--- a/skl2onnx/common/_topology.py
+++ b/skl2onnx/common/_topology.py
@@ -68,7 +68,7 @@ class Variable:
         :param type: A type object defined in .common.data_types.py;
                      e.g., FloatTensorType
         """
-        if not isinstance(raw_name, str) or '(' in raw_name:
+        if not isinstance(raw_name, str):
             raise TypeError(
                 "raw_name must be a string not '%s'." % raw_name.__class__)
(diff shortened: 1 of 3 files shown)


xadupre:
I'm ok with these changes. You should probables choose a different model to train. The training of this one generates a warning or maybe use only two variables. There is a bug in some version of scikit-learn when it goes through the code which produces this warning. Could you also check the behaviour when using ( or )?

xadupre (approved):

PR #719, opened 2021-08-27
Don't try to unreserve declared variable

### LINKED ISSUE #718: unreserve_name function does not respect already declared variables
I believe there is a bug in the `Scope.unreserve_name` function: It removes the name from the `Scope.onnx_variable_names` even if it is already declared in `Scope.variables`.

https://github.com/onnx/sklearn-onnx/blob/314cd07c19a78597c15c2ea10862b4e5d4318aa5/skl2onnx/common/_topology.py#L650-L658

If that happens then variables in `Scope.variables` could get overwritten. That is because the unique variable name is not taken anymore but a name-variable pair is already set in `Scope.variables`.

https://github.com/onnx/sklearn-onnx/blob/314cd07c19a78597c15c2ea10862b4e5d4318aa5/skl2onnx/common/_topology.py#L624-L628

This happened to me and was quite difficult to debug.

---

My suggested solution is to only unreserve an ONNX name, if the name is not already used in `Scope.variables`. If it is, then it should raise a warning instead of throwing an error. You could also make it throw an error, but then you'd have to fix it in `parse_sklearn` (which is I believe the only time the `unreserve_name` function is actually called):

https://github.com/onnx/sklearn-onnx/blob/314cd07c19a78597c15c2ea10862b4e5d4318aa5/skl2onnx/_parse.py#L566-L568

Another approach would be to get rid of `unreserve_name` altogether, since it's only called once and doesn't seem particularly useful. But maybe I'm mistaken here. 🙂

janjagusch:
Closes #718

### DIFF
--- a/skl2onnx/_parse.py
+++ b/skl2onnx/_parse.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
+import warnings
 import numpy as np
 
 from sklearn import pipeline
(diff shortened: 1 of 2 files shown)


janjagusch:
> May I close this PR since the issue was fixed in #726 ?

i will close it. thanks for your help!

PR #724, opened 2021-08-30
Fix issue when input and output have one name in common

xadupre:
Fixes issue #718. Replaces PR #719.


skl2onnx/_parse.py
```diff
@@ -559,7 +560,18 @@ def parse_sklearn(scope, model, inputs, custom_parsers=None, final_types=None):
 
         reserved = []
         for o in outputs:
-            reserved.append(scope.reserve_name(o.raw_name))
+            if scope.is_reserved(o.raw_name):
+                from logging import getLogger
+                mes = (
+                    "Name %r is reserved. It is usually due to input and "
+                    "output sharing the same name (not allowed). It will "
+                    "be renamed in %r." % (o.raw_name, o.onnx_name))
```
janjagusch, line 568:
```suggestion
                    "Name %r is reserved. It is usually due to input and "
                    "output sharing the same name (not allowed). It will "
                    "be renamed to %r." % (o.raw_name, o.onnx_name))
```

janjagusch (commented):
@xadupre, would it be possible to remove formatting changes from this PR, please? I'd like to understand better what you have done but the current diff is very large. Thanks a lot in advance!

xadupre:
> @xadupre, would it be possible to remove formatting changes from this PR, please? I'd like to understand better what you have done but the current diff is very large. Thanks a lot in advance!

Sorry for that. Some tool on Windows chose the wrong end of line by default. I'll fix it.

xadupre:
Close, replaced by #726.

PR #766, opened 2021-10-19
Add showcase of inputs being unexpectedly renamed

cbourjau:
This Draft PR adds a failing test showing the (IMHO) unexpected renaming of the input variable of the ONNX graph if it is identical to the output variable name.

### DIFF
--- a/tests/test_reproduce_regression.py
+++ b/tests/test_reproduce_regression.py
@@ -1,0 +1,63 @@
+from typing import List
+
+import numpy as np
+import pandas as pd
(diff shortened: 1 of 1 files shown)


cbourjau:
Thank you very much for taking a look at this! May I ask why the decision was made to give priority to the output names rather than the input names if the two conflict? It seems to me that having the input names as "asked for" is more important since those are mandatory at inference time (with `onnxruntime.InferenceSession.run` at least) time while the outputs are always returned as an array.

PR #777, opened 2021-11-12
Modify tfidf_transformer to enable custom vocabulary and approximate sublinear-tf scaling without sparse containers

adam444555:
The `stop_words_` attribute does not exist if custom vocabulary is provided, resulting in AttributeError. Fix it by `hasattr` check.
Approximate the sublinear_tf scaling by: first add all coefficient (included null coefficient) by 1 and then take log, i.e. replace tf with log(1+tf). Null coefficient remains to be 0 after the operation.

### DIFF
--- a/skl2onnx/operator_converters/text_vectoriser.py
+++ b/skl2onnx/operator_converters/text_vectoriser.py
@@ -226,8 +226,10 @@ def convert_sklearn_text_vectorizer(scope: Scope, operator: Operator,
             "You may raise an issue at "
             "https://github.com/onnx/sklearn-onnx/issues.")
 
-    stop_words = op.stop_words_ | (
-        set(op.stop_words) if op.stop_words else set())
+    if hasattr(op, "stop_words_"):
+        stop_words = op.stop_words_ | (set(op.stop_words) if op.stop_words else set())
+    else:
+        stop_words = set()
 
     if op.lowercase or stop_words:
         if len(operator.input_full_names) != 1:
--- a/skl2onnx/operator_converters/text_vectoriser.py
+++ b/skl2onnx/operator_converters/text_vectoriser.py
@@ -228,7 +228,7 @@ def convert_sklearn_text_vectorizer(scope: Scope, operator: Operator,
 
     if hasattr(op, "stop_words_"):
         stop_words = op.stop_words_ | (
-        set(op.stop_words) if op.stop_words else set())
+            set(op.stop_words) if op.stop_words else set())
     else:
         stop_words = set()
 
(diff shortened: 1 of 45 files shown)


skl2onnx/operator_converters/text_vectoriser.py
```diff
@@ -226,8 +226,11 @@ def convert_sklearn_text_vectorizer(scope: Scope, operator: Operator,
             "You may raise an issue at "
             "https://github.com/onnx/sklearn-onnx/issues.")
 
-    stop_words = op.stop_words_ | (
-        set(op.stop_words) if op.stop_words else set())
+    if hasattr(op, "stop_words_"):
```
xadupre, line 229: Is it possible to add a unit test to check the converter when it happens?

adam444555: A unit test is added to tests/test_sklearn_tfidf_vectorizer_converter.py, which accept a custom vocabulary as an additional parameter to TfidfVectorizer. Besides that, is there any modfication needed to adopt the unit test?

tests/test_sklearn_tfidf_vectorizer_converter.py
```diff
@@ -577,6 +577,33 @@ def test_model_tfidf_vectorizer_nan(self):
...
+        corpus = numpy.array([
+            "This is the first document.",
+            "This document is the second document.",
+            "And this is the third one.",
+            "Is this the first document?",
+        ]).reshape((4, 1))
+        vc = ["first", "second", "third", "document", "this"]
+        vect = TfidfVectorizer(ngram_range=(1, 1), norm=None, vocabulary=vc)
+        vect.fit(corpus.ravel())
+        model_onnx = convert_sklearn(vect, "TfidfVectorizer",
```
xadupre, line 593: Can you make sure that attribute `stop_words_` is missing?

adam444555: Yes, here is the description of stop_words_ attributes from sklearn:
`    stop_words_ : set
        Terms that were ignored because they either:
          - occurred in too many documents (`max_df`)
          - occurred in too few documents (`min_df`)
          - were cut off by feature selection (`max_features`).
        This is only available if no vocabulary was given.`
And this is what I got before modified:
`AttributeError: 'TfidfVectorizer' object has no attribute 'stop_words_'`

xadupre:
https://github.com/onnx/sklearn-onnx/pull/777#discussion_r748557402: sorry I was not clear enough, I just meant adding something like `assert not hasattr(vect, 'stop_words_')`.

xadupre (approved):

adam444555:
A specific commit could not be signed off (I tried with many approaches but no one worked, and I had no idea why). Therefore I rebase the branch to drop the problemetic commit, and then recommit the same change.

xadupre (approved):

xadupre:
Sorry for the trouble. Let me know if it is ready to be merged.

adam444555:
> Sorry for the trouble. Let me know if it is ready to be merged.

Yes, it is ready right now!

PR #810, opened 2022-01-14
Documentation: Ways to create own converters.

JulZimmermann:
The documentation was inconsistent with the mentioned order of the two ways to create a converter.


xadupre:
Thanks for doing it. You need to follow the instructions here: https://github.com/onnx/sklearn-onnx/pull/810/checks?check_run_id=4813891653 to make failing test pass. For further commit, use `git commit -s ...`.

xadupre (approved):

PR #824, opened 2022-02-09
Fixes for onnx==1.11.0

xadupre:
* remove many lines of codes introduced in the early stages of the development and not needed anymore
* introduce TARGET_OPSET_ML for testing purpose
* use TARGET_OPSET to enable, disable unittest instead of onnx.__version__


BowenBao:
Thanks @xadupre for doing this check, I can follow up on this if you want. 
The CI for onnx 1.11-rc seems to need installing `wget` beforehand.

xadupre:
> Thanks @xadupre for doing this check, I can follow up on this if you want. The CI for onnx 1.11-rc seems to need installing `wget` beforehand.

It seems so. That's a surprising dependency.

tests/test_algebra_complex.py
```diff
@@ -21,8 +19,7 @@ class TestAlgebraComplex(unittest.TestCase):
 
     @unittest.skipIf(Complex64TensorType is None,
                      reason="not available")
-    @unittest.skipIf(StrictVersion(onnx.__version__) < StrictVersion('1.8.0'),
-                     reason="not implemented")
+    @unittest.skipIf(TARGET_OPSET < 13, reason="not implemented")
```
BowenBao, line 22: Similar to above should it be  ` <14`？

tests/test_algebra_converters.py
```diff
@@ -18,8 +16,7 @@
 
 class TestAlgebraConverters(unittest.TestCase):
 
-    @unittest.skipIf(StrictVersion(onnx.__version__) < StrictVersion("1.4.0"),
-                     reason="not available")
+    @unittest.skipIf(TARGET_OPSET < 10, reason="not available")
```
BowenBao, line 19: ditto

tests/test_sklearn_tfidf_vectorizer_converter.py
```diff
@@ -22,22 +22,22 @@
...
 from skl2onnx import convert_sklearn
 from skl2onnx.common.data_types import StringTensorType, FloatTensorType
-import onnx
-import onnxruntime
+from onnxruntime import __version__ as ort_version
 from onnxruntime import InferenceSession
 from test_utils import dump_data_and_model, TARGET_OPSET
 
 
+ort_version = '.'.join(ort_version.split('.')[:2])
```
BowenBao, line 30: nit: this appears multiple time, maybe put into test_utils?

tests/test_algebra_cascade.py
```diff
@@ -25,8 +23,7 @@
 
 class TestOnnxOperatorsCascade(unittest.TestCase):
 
-    @unittest.skipIf(StrictVersion(onnx.__version__) < StrictVersion("1.4.0"),
```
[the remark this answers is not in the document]
BowenBao: update: after looking at more changes, it seems the new opset version should be more accurate. Feel free to resolve if that's the case.

BowenBao (approved):
Thanks for the updates. Looks good. Left a few questions checking if my understanding is correct.

tests/test_sklearn_adaboost_converter.py
```diff
@@ -72,21 +61,13 @@ def test_ada_boost_classifier_samme_r_decision_function(self):
         )
         self.assertIsNotNone(model_onnx)
         dump_data_and_model(
-            X_test,
+            X_test[:5],
```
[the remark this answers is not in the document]
xadupre: I forgot to remove it. I did it after adding verbose=True to understand the failure.

tests/test_sklearn_tfidf_vectorizer_converter.py
xadupre: Then the code can work with onnxruntime-training as well.

tests/test_algebra_complex.py
xadupre: See https://github.com/onnx/onnx/blob/main/docs/Versioning.md.

PR #826, opened 2022-02-12
Enable Double input type for RandomForest and ZipMap

BowenBao:
* Regardless of input type, attribute types remains `Float` for floating attributes of RandomForest.
* Regardless of input type, probabilities output type of RandomForest remains `Float`.
* `ZipMap` only supports `Float` input & output type. Update shape_calculator to infer output type from input type.

Signed-off-by: BowenBao <bowbao@microsoft.com>

### DIFF
--- a/skl2onnx/_parse.py
+++ b/skl2onnx/_parse.py
@@ -13,7 +13,8 @@ except ImportError:
     class OutlierMixin:
         pass
 
-from sklearn.ensemble import IsolationForest, RandomTreesEmbedding, RandomForestClassifier
+from sklearn.ensemble import (
+    IsolationForest, RandomTreesEmbedding, RandomForestClassifier)
 from sklearn.gaussian_process import GaussianProcessRegressor
 from sklearn.linear_model import BayesianRidge
 from sklearn.model_selection import GridSearchCV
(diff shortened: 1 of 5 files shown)


xadupre:
Until ml opset 1, float attributes were the only supported ones (see https://github.com/onnx/onnx/blob/main/docs/Changelog-ml.md#ai.onnx.ml.TreeEnsembleClassifier-1). It changes in ml opset 3 (https://github.com/onnx/onnx/blob/main/docs/Operators-ml.md#ai.onnx.ml.TreeEnsembleClassifier). This was introduced to fix discrepancies due to the float attributes (see https://onnx.ai/sklearn-onnx/auto_tutorial/plot_ebegin_float_double.html).

BowenBao:
@xadupre Thanks for info. Hence for opset 3, I only need to update the output type of `TreeEnsembleClassifier`, and type for `ZipMap`. Attribute type can remain double, unless it's opset <= 2. 

ONNXRuntime does not yet support opset 3, so customer needs to specify onnx ml opset 2 for export. Is that achievable through `target_opset` argument in `convert_sklearn`? The description says `number, for example, 7 for ONNX 1.2, and 8 for ONNX 1.3,` which seems like onnx opset but not onnx ml opset.

xadupre (approved):

xadupre:
The build is fixed in PR #827 (an example in the documentation refers to an onnx model on github and its location has changed).

PR #844, opened 2022-03-15
fix minor bug and refine doc

xiaowuhu:
fix minor bug
Signed-off-by: xiaowuhu <hu.xiaowu@hotmail.com>

### DIFF
--- a/docs/tutorial/plot_bbegin_measure_time.py
+++ b/docs/tutorial/plot_bbegin_measure_time.py
@@ -72,7 +72,7 @@ with config_context(assume_finite=True):
         obs.append(mt)
 
 df_skl = DataFrame(obs)
-df_skl
+print(df_skl)
 
 #####################################
 # Graphe.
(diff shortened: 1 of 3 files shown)


docs/tutorial/plot_bbegin_measure_time.py
```diff
@@ -122,7 +122,7 @@
 
 
 df = DataFrame(obs)
-df
+print(df)
```
xadupre, line 125: I let it that that because the rendering is better in the documentation.

xadupre (approved):

docs/tutorial/plot_bbegin_measure_time.py
xiaowuhu: got it. I reversed this change.

xiaowuhu: got it. I reversed this change.

xadupre (approved):

PR #855, opened 2022-05-04
MAINT compat link function and loss for sklearn 1.1

glemaitre:
We released an RC for scikit-learn 1.1.

I tried to check if we break anything in `sklearn-onnx`.
I could spot a couple of issues linked to some change of a default for the loss in GBDT and we internally refactor the losses and link functions in the GLM.

I am not familiar with the code so the tests might still not pass but it could be a good approximation to show what to do to migrate to scikit-learn 1.1

@xadupre might be interested to have a look.


xadupre:
Thanks for the contribution. I'll look into that this week.

xadupre:
Closing the PR as it was merged by #863.

PR #904, opened 2022-08-30
Xiaowu/one vs one classifier

xiaowuhu:
add OneVsOneClassifier converter.


skl2onnx/operator_converters/one_vs_one_classifier.py
```diff
@@ -0,0 +1,175 @@
...
+    op = operator.raw_operator
+    options = container.get_options(op, dict(raw_scores=False))
+    use_raw_scores = options['raw_scores']
+    label_names = []
+    score_names = []
+    for i, estimator in enumerate(op.estimators_):
+        op_type = sklearn_operator_name_map[type(estimator)]
+        label_name_out = 'label_out_%d' % i
+        score_name_out = 'score_out_%d' % i
+        if op_type == 'SklearnLinearSVC':
```
xadupre, line 33: You should something like that: https://github.com/onnx/sklearn-onnx/blob/main/skl2onnx/operator_converters/one_vs_rest_classifier.py#L37.

skl2onnx/operator_converters/one_vs_one_classifier.py
```diff
@@ -0,0 +1,175 @@
...
+
+    n_classes = len(op.classes_)
+    sumc_name = [scope.get_unique_variable_name('svcsumc_%d' % i)
+                 for i in range(n_classes)]
+    vote_name = [scope.get_unique_variable_name('svcvote_%d' % i)
+                 for i in range(n_classes)]
+    sumc_add = {n: [] for n in sumc_name}
+    vote_add = {n: [] for n in vote_name}
+    k = 0
+    for i in range(n_classes):
```
xadupre, line 93: You should use this operator: https://github.com/onnx/sklearn-onnx/blob/main/skl2onnx/operator_converters/ovr_decision_function.py#L154. The code is the same.

skl2onnx/operator_converters/one_vs_one_classifier.py
```diff
@@ -0,0 +1,84 @@
...
+            p1 = scope.get_unique_variable_name('probY_%d' % i)
+            apply_slice(scope, prob_name.onnx_name, p1, container, starts=[1],
+                        ends=[2], axes=[1],
+                        operator_name=scope.get_unique_operator_name('Slice'))
+
+        probs_names.append(p1)
+
+
+    this_operator = scope.declare_local_operator("SklearnOVRDecisionFunction", op)
+    this_operator.inputs.append(probs_names)
```
xadupre, line 68: I think you need to add a node 'Concat' and make it the single input of pseudo operator SklearnOVRDecisionFunction.

xiaowuhu:
@xadupre it almost work, but the score has not exactly same with expectation, I guess something not correct in the  ovr_decision_function().

xadupre:
> @xadupre it almost work for the label, but the scores are not exactly same with expectation, I guess something not correct in the ovr_decision_function().

ovr is right but it uses probabilities when available, raw_scores otherwise. I need to change that.

requirements.txt
```diff
@@ -4,3 +4,4 @@ protobuf
 onnx>=1.2.1
 scikit-learn>=0.19
 onnxconverter-common>=1.7.0
+scikit-learn<=1.1.1
```
xadupre, line 7: not sure this one is needed since you fixed the weird issue by using `sudo apt-get upgrade libstdc++6`.

xadupre (approved):

PR #930, opened 2022-10-30
Added support for passthrough connection to stacking estimator converter

OKUA1:
As described in sklearn documentation, both StackingClassifier and StackingRegressor accept a `passthrough` argument, which (when set to True) allows to pass the original input data to the final estimator creating something similar to a skip connection. 

This functionality is not realized in current converter which makes a converted model invalid in cases when `passthrough` was set to True during training.

### DIFF
--- a/tests/test_sklearn_stacking.py
+++ b/tests/test_sklearn_stacking.py
@@ -35,21 +35,23 @@ from test_utils import (
     fit_classification_model, TARGET_OPSET)
 
 
-def model_to_test_reg():
+def model_to_test_reg(passthrough=False):
     estimators = [
         ('dt', DecisionTreeRegressor()),
         ('las', LinearRegression())]
     stacking_regressor = StackingRegressor(
-        estimators=estimators, final_estimator=LinearRegression())
+        estimators=estimators, final_estimator=LinearRegression(),
+        passthrough=passthrough)
     return stacking_regressor
 
 
-def model_to_test_cl():
+def model_to_test_cl(passthrough=False):
     estimators = [
         ('dt', DecisionTreeClassifier()),
         ('las', LogisticRegression())]
     stacking_regressor = StackingClassifier(
-        estimators=estimators, final_estimator=LogisticRegression())
+        estimators=estimators, final_estimator=LogisticRegression(),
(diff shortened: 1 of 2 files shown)


xadupre:
Is it possible to add a unit test for it?

OKUA1:
@xadupre Yes, working on it right now. Quick question: when testing the regressor, some of the predictions are above the threshold causing the test to fail (the unit test for classification passes without issues). Could it be caused simply by float64->float32 conversion or should I look for another cause? And if so, do you have some advice how to properly debug it? 

The test output:

```
--expected--output--

Arrays are not almost equal to 5 decimals

Mismatched elements: 11 / 125 (8.8%)
Max absolute difference: 2.32062132e-05
Max relative difference: 4.41075741e-06
onnx==1.12.0 onnxruntime==1.12.1 TARGET_OPSET=17
onnx==1.12.0 onnxruntime==1.12.1 TARGET_OPSET=17
onnx==1.12.0 onnxruntime==1.12.1 TARGET_OPSET=17

----------------------------------------------------------------------
```

OKUA1:
I added several unit tests which match current ones, but with `passthrough = True`. 

As mentioned earlier, the regression was/is a bit problematic one with deviation in the predictions of onnx/sklearn being slightly above the desired threshold for a small subset of tested samples. After further investigation I did not get any ideas what the cause could be except the originally assumed discrepancy due to the usage of float32 in onnx.  For now, I used `factor=0.1` in the respective unit test which prevents it from failing but probably not the best thing to do.

tests/test_sklearn_stacking.py
```diff
@@ -270,6 +351,83 @@ def converter(scope, operator, container):
         got = sess.run(None, {'X': x})[0]
         self.assertEqual(got.shape[0], x.shape[0])
 
+    @unittest.skipIf(StackingClassifier is None,
+                     reason="new in 0.22")
+    @ignore_warnings(category=FutureWarning)
+    def test_concat_stacking_passthrough(self):
+
+        class CustomTransformer:
```
xadupre, line 359: Just curious, why not inheriting from Estimator, TransformerMixin instead of implementing a new parser?

PR #966, opened 2023-02-14
Concat | Shape Calculator 

fillassuncao:
Signed-off-by: fillassuncao <fillassuncao@gmail.com>

This PR addresses some issues in the shape calculator of the concat operator, namely:
 - when computing the `seen_types` we are only adding when the list is empty;
 - the computation of the `final_type` is not correct.

### DIFF
--- a/skl2onnx/shape_calculators/concat.py
+++ b/skl2onnx/shape_calculators/concat.py
@@ -32,7 +32,7 @@ def calculate_sklearn_concat(operator):
                 C += 1
             else:
                 C = None
-        if len(seen_types) == 0:
+        if i.type not in seen_types:
             seen_types.append(i.type)
 
     def more_generic(t1, t2):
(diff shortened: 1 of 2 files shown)


xadupre:
Is it possible to add a unit test with an example for which the previous implementation did not work?

fillassuncao:
> Is it possible to add a unit test with an example for which the previous implementation did not work?

Sure can, will just need to allocate time for this. While I am doing it, I will next provide an example where the current implementation breaks.

Imagine a scenario where there are two inputs to the concat operator, the first is an `int` and the second a `float`.  With the current implementation, the typing of the output of the concat operator will be set to `int`.

Later on, because the expected type is `int`, all inputs to the concat operator will be cast to `int` (instead of `float`) (casts are done in [here](https://github.com/onnx/sklearn-onnx/blob/bbe2f6d724e0bda44c5a1147b92e05686625aaa0/skl2onnx/operator_converters/concat_op.py#L18))

fillassuncao:
@xadupre just added a unit test -  `TestConcatOutputType`

Issue #969, opened 2023-03-02
Large differences between scikit-learn GaussianProcessRegressor with Matern() kernel and ONNX model


whophil:
I am seeing an issue where a scikit-learn GaussianProcessRegressor whose kernel contains `Matern()` and its ONNX-converted model give drastically different prediction results. I am using `DoubleTensorType`, following the instructions of http://onnx.ai/sklearn-onnx/auto_examples/plot_gpr.html#third-attempt-use-of-double

Below is some code demonstrating the issue

<details>
<summary>Example code</summary>

```python
from sklearn.datasets import load_boston
from sklearn.gaussian_process import GaussianProcessRegressor
from sklearn.gaussian_process.kernels import DotProduct, RBF, ConstantKernel, Matern, WhiteKernel
from sklearn.model_selection import train_test_split
import onnxruntime as rt
from skl2onnx.common.data_types import DoubleTensorType
from skl2onnx import convert_sklearn


def convert_skl_gpr(gpr):
    initial_type = [('X', DoubleTensorType([None, None]))]
    onx = convert_sklearn(gpr, initial_types=initial_type,
                            target_opset=12)
    return onx


def evaluate_onx(onx, X):
    sess64 = rt.InferenceSession(onx.SerializeToString())
    return sess64.run(None, {'X': X})[0].squeeze()


# load, split data
bost = load_boston()
X, y = bost.data, bost.target
X_train, X_test, y_train, y_test = train_test_split(X, y)
X_test = X_test[:5]

# GPR with DotProduct + RBF
gpr_1 = GaussianProcessRegressor(DotProduct() + RBF(), alpha=1.)
gpr_1.fit(X_train, y_train)

pred_skl = gpr_1.predict(X_test)
pred_onx = evaluate_onx(convert_skl_gpr(gpr_1), X_test)
print('DotProduct + RBF')
print('-' * 80)
print('skl', pred_skl)
print('onx', pred_onx)
print('diff', pred_skl - pred_onx)

# GPR with Matern
gpr_2 = GaussianProcessRegressor(Matern(), alpha=1.)
gpr_2.fit(X_train, y_train)

pred_skl = gpr_2.predict(X_test)
pred_onx = evaluate_onx(convert_skl_gpr(gpr_2), X_test)
print('Matern')
print('-' * 80)
print('skl', pred_skl)
print('onx', pred_onx)
print('diff', pred_skl - pred_onx)
```

</details>

Which outputs
```
DotProduct + RBF
[...]

PR #978, opened 2023-03-15
Add kernel Matern for gaussian process

xadupre:
Reply to issue #969.


whophil:
Thanks @xadupre for starting on this! Just wondering if this is something you are actively developing, or if there are some other blocking issues?

PR #986, opened 2023-04-07
Swayand/bug fix min max scaler with clip

swayand:
Hi,

there is a bug when using the MinMaxScaler. By looking at the comment in:
https://github.com/onnx/sklearn-onnx/blob/3c9e0c1a757229f63b94f01cef3a51b638f91f54/skl2onnx/operator_converters/scaler_op.py#L210
the solution is was easy ;) 

In the first commit i expand the test, so that the bug can be reproduced. The short summary, min max values can be different than the feature_range (unfortunately in the orginal test the values where the same). 

In the second commit I fixed the problem, which also makes the code more readable.


xadupre (approved):

swayand:
Hi @xadupre,

from my side the fix is complete ;)

xadupre (approved):

PR #1011, opened 2023-08-07
Update Version to 1.15

xiaowuhu:
- update version to 1.15


xiaowuhu:
@xadupre please help to review.

.pypirc
```diff
@@ -0,0 +1,16 @@
+# This is template for .pypirc file
```
xadupre, line 1: I don't think we should include this file.

xiaowuhu: removed

xadupre (approved):

Issue #1014, opened 2023-08-12
Use onnx graph from PyTorch `torch.onnx.export` in custom converter


KarelZe:
Thanks for your effort on slearn2onnx. 👍 

**Summary**
I have PyTorch models wrapped as a sklearn estimator in my sklearn pipeline. I want to use the onnx graph derived from `torch.onnx.export` as a custom converter for conversion.

**Comprehensive**

My question is related to https://github.com/onnx/sklearn-onnx/issues/597. I have implemented an AutoEncoder in PyTorch, from which I use the encoder for dimensionality reduction. The autoencoder is wrapped in sklearn-compatible wrapper named `AutoEncoderModel`. Similar concept exists with libraries like [`skorch`](https://github.com/skorch-dev/skorch). The wrapper derives from `BaseEstimator`, `TransformerMixin`, `ClassNamePrefixFeaturesOutMixin` . The `fit()` in the `AutoEncoderModel` trains the Autoencoder and `transform()` passes the input through the encoder.

I'm using the following [PyTorch code](https://pytorch.org/docs/stable/onnx.html) to generate the onnx graph for the encoder inside my wrapper `AutoEncoderModel`:
```python
input_names = ["input"] + [
            layer.__class__.__name__ for layer in self.module.encoder[:-1]
]

cache = io.BytesIO()
torch.onnx.export(
            self.module.encoder,
            torch.ones(size=(self.batch_size, n_features_in)),
            cache,
            input_names=input_names,
            output_names=["bottleneck"],
            keep_initializers_as_inputs=False,
            verbose=True,
            opset_version=15,
)
onx = onnx.load_model_from_string(cache.getvalue())
```
My Autoencoder / AutoEncoder/Model has the following structure:
```python
AutoEncoderModel(epochs=2,
                 module=AutoEncoder(
  (encoder): Sequential(
    (0): Linear(in_features=50, out_features=30, bias=True)
    (1): ReLU()
    (2): Linear(in_features=30, out_features=20, bias=True)
  )
  (decoder): Sequential(
    (0): Linear(in_features=20, out_features=30, bias=True)
    (1): ReLU()
    (2): Linear(in_features=30, out_features=50, bias=True)
  )
))
```
The resulting graph looks like this:
```proto
ir_version: 8
opset_import {
  version: 15
}
producer_name: "pytorch"
producer_version: "2.0.1"
graph {
  node {
    input: "input"
    input: "Linear"
    input: "ReLU"
[...]

PR #1023, opened 2023-08-30
Add function add_onnx_graph to insert an onnx graph in a pipeline

xadupre:
Function add_onnx_graph makes it easier to write custom converters for model coming from other framework. See issue #1014 for an example.


KarelZe:
Thanks for your work @xadupre. Is it possible to make it accessible in the [`__init__.py`](https://github.com/onnx/sklearn-onnx/blob/main/skl2onnx/helpers/__init__.py)?

xadupre:
Like from skl2onnx.helpers import add_onnx_graph? I'll do that.

Issue #998, opened 2023-06-07
Skipping out-of-vocabulary n-grams during batch inference of Tf-IDF ONNXified Vectorizer 


nssprogrammer:
An ONNXified TF-IDF vectorizer while during batch inference is dropping out-of-vocabulary texts whereas the expected behaviour is  it should return zero vectors for those out-of-vocabulary texts.

Example :- 
Suppose a batch contains 5 documents [ d1 , d2 , d3 , d4 , d5 ] , let d3 and d5 are out-of-vocabulary texts , so the expected  output should be ideally [ v1 , v2 , 0 , v4 , 0 ] , returning corresponding vectors. But the output the model returns is - [ v1 , v2 , v4 ] 

ONNX v7
skl2onnx 1.13

xadupre:
Is there a tiny model I can use to replicate your issue and see what I can do to fix it?

vsbaldeev:
@xadupre code mentioned in this [ussue](https://github.com/microsoft/onnxruntime/issues/17795) to reproduce this bug

`python
from sklearn.feature_extraction.text import CountVectorizer
from skl2onnx import convert_sklearn
from skl2onnx.common.data_types import StringTensorType
from sklearn.pipeline import Pipeline
from sklearn.linear_model import LogisticRegression
import onnxruntime

corpus = [
    'This is the first document.',
    'This document is the second document.',
    'And this is the third one.',
    'Is this the first document?',
]

labels = [
    'a',
    'b',
    'c',
    'd'
]

to_inference = [
    '.',
    'first'
]

vectorizer = CountVectorizer(
    max_features=5,
    analyzer='word',
    ngram_range=(1, 1),
    encoding='utf8',
    strip_accents=None,
    token_pattern=(
            r"\b[a-zA-Z0-9_]+\b"
            r"|"
            r"[\~\-!.,:;@+&<>*={}\[\]№?()^|/%$#'`\"\\_]"
            r"|"
            r"\d+"
        )
)
vectorizer.fit_transform(corpus)
print(vectorizer.get_feature_names_out())


classifier = LogisticRegression()


model = Pipeline(steps=[("vectorizer", vectorizer), ("classifier", classifier)])
model.fit(corpus, labels)
print(model.predict(to_inference))

onnx_model = convert_sklearn(model, initial_types=[('X', StringTensorType((None,)))])


session = onnxruntime.InferenceSession(onnx_model.SerializeToString(), providers=['CPUExecutionProvider'])
[...]

PR #1031, opened 2023-10-05
StringNormalizer drops strings when they only contain stop words

xadupre:
Issue raised in https://github.com/microsoft/onnxruntime/issues/17795 and https://github.com/onnx/sklearn-onnx/issues/998. After investigation, operator StringNormalizer drops strings containing only stopwords. The expected output can be obtained by running the inference for every input independantly.

A possible fix would be to update onnx specifications to not drop strings. Another fix could be to update the converter and loop over rows but that's not very efficient.


vsbaldeev:
@xadupre Hi! Any updates on this problem ?

As I understand the fast fix would be to process every string independently. RIght ?

xadupre:
That would work.

xadupre:
If you are using a loop, it is not really suprising. There is no parallelization even though each row is processed independently.

vsbaldeev:
> If you are using a loop, it is not really suprising. There is no parallelization even though each row is processed independently.

May you suggest some impovement in this case ?

My guess is if I know indices of dropped strings then I can rerun inference session only on them instead loop over all input items.

vsbaldeev:
Also I want to notice that I don't user stop_words parameter. How exactly set stop_words parameter to empty or None while converting to ONNX model ?

xadupre:
One issue is StringNormalizer is defined in onnx. To change its behaviour, it has to be changed in onnx and onnxruntime. It is long. onnxruntime-extensions is a project implementing custom kernel without going through the process of validating the definition by onnx mainteners. The best option here is to implement is to implement a custom kernel. I'm hesitating in adding more in onnxruntime or creating a similar projet to host custom kernel dedicated to standard machine learning. But a custom kernel would solve the issue.

vsbaldeev:
@xadupre What do you mean by custom kernel exactly ?

xadupre:
A C++ implementation of an operator. Let's assume we define the operator StringNormalizer2. onnxruntime does not know it because it is new so it cannot run the inference unless we give it a piece of code compiled into an assembly. onnxruntime lets you do that. See https://onnxruntime.ai/docs/reference/operators/add-custom-op.html.

xadupre:
This issue seems to be fixed with scikit-learn==1.5.0. Could you check on your side?

PR #1041, opened 2023-11-07
Add custom converter for ClassificationProgressiveLearner 

certh-knowledge-project:
Hi,

This pull request implements a custom converter regarding the [`ClassificationProgressiveLearner`](https://github.com/neurodata/ProgLearn/blob/03ab355c10e58c41c9023996a48c39937414ec9a/proglearn/progressive_learner.py#L732) estimator as is defined in the neurodata [proglearn repository](https://github.com/neurodata/ProgLearn). This package is designed for exploring and using Progressive learning algorithms. More specifically, Progressive learning is a concept in machine learning where the model learns from a series of tasks progressively, leveraging the knowledge gained from previous tasks to improve performance on later tasks. This approach can be particularly useful in scenarios where data comes in a sequential manner or when it’s beneficial to transfer knowledge from one task to another. If you are unfamiliar with the concept or wish to explore it in greater detail, you can find more information in the accompanying [paper](https://arxiv.org/abs/2004.12908) by Jayanta Dey et al.

#### PR Details
The Progressive learning implementation is split into two separate categories, [`TreeClassificationTransformer`](https://github.com/neurodata/ProgLearn/blob/03ab355c10e58c41c9023996a48c39937414ec9a/proglearn/transformers.py#L133) and [`NeuralClassificationTransformer`](https://github.com/neurodata/ProgLearn/blob/03ab355c10e58c41c9023996a48c39937414ec9a/proglearn/transformers.py#L13), each utilizing a different underlying estimator. `TreeClassificationTransformer` is based on an ensemble of `scikit-learn` [`DecisionTreeClassifier`](https://scikit-learn.org/stable/modules/generated/sklearn.tree.DecisionTreeClassifier.html) instances (converters already implemented in `skl2onnx`) whereas the latter is centered around `tensorflow/keras` neural networks. Since `skl2onnx` solely focuses on `scikit-learn` estimators, this contribution emphasizes only on `TreeClassificationTransformer` to avoid conflicts. Concretely, this contribution is loosely based on the [`pyod.models.iforest.IForest`](http://onnx.ai/sklearn-onnx/auto_tutorial/plot_wext_pyod_forest.html) custom converter, following its main set of instructions adapted to our needs. 

#### Requirements
In an effort to keep the core `skl2onnx` dependencies pure, the original `requirements.txt` file is not modified, however, as a note the necessary additional dependencies are presented below (excluding the ones already required by `skl2onnx`).

```sh
scikit-learn>=1.0
tensorflow>=2.4.0
proglearn>=0.0.7
```

#### Testing
The main entry point should be the **`test_proglearn.py`** file, inside the **`tests`** folder, which defines a total of **7 independent test cases**, after registering the external converters/shape calculators. It should be noted at this point that a version of `onnxruntime>=1.8.0` is required to perform these tests.

_Note_: There is a known backtracking issue with pip which may result in multiple versions of the packages being installed. A quick fix is to use the `--use-deprecated legacy-resolver` option when running `pip install`.

Thank you for taking the time to review this pull request! Looking forward to your insights and suggestions.


xadupre:
I would prefer to see these converters in a separate package. Taking a dependency on tensorflow and proglearn is a huge change. I assume more converters would be added in the future. I'm reluctant to start fixing the converters when they fail because a new version of proglearn was released. sklearn-onnx has an API to register custom converters. Creating a separate package should not be an issue.

certh-knowledge-project:
Hi @xadupre,

Thank you for taking the time to review the pull request and for your valuable feedback. We understand your concerns about adding a dependency on like TensorFlow and ProgLearn, and the potential issues that could arise with future updates. 

Your suggestion to create a separate package for the converters is a good one, and we appreciate your insight about the API in sklearn-onnx to register custom converters. We will take your feedback into consideration as we continue to improve our project. Thanks again for your thoughtful review.

Issue #609, opened 2021-02-05
FunctionTransformer is not supported


HamzaSaouli:
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-166-79eb1b0d84fa> in <module>
      1 # Export the model
      2 initial_type = [('numfeat', FloatTensorType([None, 30]))] 
----> 3 model_onnx = convert_sklearn(model, initial_types=initial_type)
      4 
      5 # Save it into wanted file

~/.local/lib/python3.6/site-packages/skl2onnx/convert.py in convert_sklearn(model, name, initial_types, doc_string, target_opset, custom_conversion_functions, custom_shape_calculators, custom_parsers, options, dtype, intermediate, white_op, black_op, final_types)
    148 
    149     # Infer variable shapes
--> 150     topology.compile()
    151 
    152     # Convert our Topology object into ONNX. The outcome is an ONNX model.

~/.local/lib/python3.6/site-packages/skl2onnx/common/_topology.py in compile(self)
    901         self._resolve_duplicates()
    902         self._fix_shapes()
--> 903         self._infer_all_types()
    904         self._check_structure()
    905 

~/.local/lib/python3.6/site-packages/skl2onnx/common/_topology.py in _infer_all_types(self)
    754                 shape_calc(operator)
    755             else:
--> 756                 operator.infer_types()
    757 
    758     def _resolve_duplicates(self):

~/.local/lib/python3.6/site-packages/skl2onnx/common/_topology.py in infer_types(self)
    220                 "Unable to find a shape calculator for alias '{}' "
    221                 "and type '{}'.".format(self.type, type(self.raw_operator)))
--> 222         shape_calc(self)
    223 
    224     @property

~/.local/lib/python3.6/site-packages/skl2onnx/shape_calculators/function_transformer.py in calculate_sklearn_function_transformer_output_shapes(operator)
     15     """
     16     if operator.raw_operator.func is not None:
---> 17         raise RuntimeError("FunctionTransformer is not supported unless the "
     18                            "transform function is None (= identity). "
     19                            "You may raise an issue at "

RuntimeError: FunctionTransformer is not supported unless the transform function is None (= identity). You may raise an issue at https://github.com/onnx/sklearn-onnx/issues.

xadupre:
[...]

PR #1042, opened 2023-11-08
Adds an example about FunctionTransformer

xadupre:
Answer to issue #609.


docs/tutorial/plot_function_transformer.py
```diff
@@ -0,0 +1,193 @@
...
+        pass
+
+    def calculate_overprice(self, x, y):
+        return 100 * (x - y) / y
+
+    def fit(self, X, y=None):
+        return self
+
+    def transform(self, X, y=None):
+        x = X.apply(lambda x: self.calculate_overprice(x.a, x.b), axis=1)
```
paranjapeved15, line 85: I would like to return all other columns (a,b,f) in the data frame along with the new column "c"

docs/tutorial/plot_function_transformer.py
```diff
@@ -0,0 +1,193 @@
...
+# The first function returns the expected type and shape.
+
+
+def overprice_shape_calculator(operator):
+    check_input_and_output_numbers(operator, input_count_range=1, output_count_range=1)
+    # Gets the input type, the transformer works on any numerical type.
+    input_type = operator.inputs[0].type.__class__
+    # The first dimension is usually dynamic (batch dimension).
+    input_dim = operator.inputs[0].get_first_dimension()
+    operator.outputs[0].type = input_type([input_dim, 1])
```
paranjapeved15, line 127: to return all 4 columns, something needs to change in the operator outputs shape, correct?

docs/tutorial/plot_function_transformer.py
```diff
@@ -0,0 +1,193 @@
...
+
+    # 100 * (x-y)/y  --> 100 * (X[0] - X[1]) / X[1]
+
+    zero = np.array([0], dtype=np.int64)
+    one = np.array([1], dtype=np.int64)
+    two = np.array([2], dtype=np.int64)
+    hundred = np.array([100], dtype=np.float32)
+
+    # Slice(data, starts, ends, axes)
+    x0 = OnnxSlice(X, zero, one, one, op_version=opv)
```
paranjapeved15, line 144: I feel like the learning curve to convert normal operations to onnx operations is pretty steep. Could you suggest any good tutorials / docs to get more familiar with this?

docs/tutorial/plot_function_transformer.py
xadupre: I updated the pipeline to retain those two: `("ab", FunctionTransformer(), ["a", "b"]),`

docs/tutorial/plot_function_transformer.py
xadupre: No, this transformer takes 2 inputs and produces one. I added the others in other piece in the pipeline.

docs/tutorial/plot_function_transformer.py
```diff
@@ -0,0 +1,206 @@
...
+def overprice_shape_calculator(operator):
+    check_input_and_output_numbers(operator, input_count_range=1, output_count_range=1)
+    # Gets the input type, the transformer works on any numerical type.
+    input_type = operator.inputs[0].type.__class__
+    # The first dimension is usually dynamic (batch dimension).
+    input_dim = operator.inputs[0].get_first_dimension()
+    operator.outputs[0].type = input_type([input_dim, 1])
+
+
+def overprice_converter(scope, operator, container):
```
paranjapeved15, line 135: could you also please replace all occurrences of the word "overprice" with something else? I forgot to rename it while posting

docs/tutorial/plot_function_transformer.py
xadupre: For this example, I chose the API implemented in sklearn-onnx but I could extend the example to make it work with other API you prefer (see [Many ways to implement a custom graph in ONNX](https://sdpython.github.io/doc/onnx-array-api/dev/tutorial/onnx_api.html)) to write onnx graphs.

That said, I understand it is not easy. It takes time to rewrite an existing code with this list of [ONNX Operators](https://onnx.ai/onnx/operators/).

The easiest way is probably to use pytorch but it means to take a dependency on pytorch. The numpy API for onnx is less difficult but I initially started it for small functions. All the other options require to rewrite a function with onnx operators. Most of the operators follow numpy api. A couple of important points: there is no inplace modification allowed in onnx, notation [..., :, ...] can usually be handled with operator Slice but not always, Loop and Tests should be avoided.

I wrote this tutorial about onnx: https://onnx.ai/onnx/intro/python.html.

paranjapeved15: Why is using PyTorch easier than sklearn for converting to onnx?

xadupre: It is not necessary to rewrite your custom code with pytorch. With sklearn, the numpy code must be rewritten. I added a version with onnxscript as well.

docs/tutorial_2_new_converter.rst
```diff
@@ -42,6 +42,7 @@ an example.
 
     auto_tutorial/plot_icustom_converter
     auto_tutorial/plot_jcustom_syntax
+    auto_tutorial/plot_function_transformer
```
fatcat-z, line 45: Nit:
Move it up so the names are alphabetically.

xadupre: I renamed it to keep that order. It makes more sense to keep it at this position.

paranjapeved15 (approved):

docs/tutorial/plot_jfunction_transformer.py
```diff
@@ -0,0 +1,292 @@
...
+def growth_shape_calculator(operator):
+    check_input_and_output_numbers(operator, input_count_range=1, output_count_range=1)
+    # Gets the input type, the transformer works on any numerical type.
+    input_type = operator.inputs[0].type.__class__
+    # The first dimension is usually dynamic (batch dimension).
+    input_dim = operator.inputs[0].get_first_dimension()
+    operator.outputs[0].type = input_type([input_dim, 1])
+
+
+def growth_converter(scope, operator, container):
```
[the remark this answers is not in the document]
paranjapeved15: Thanks @xadupre . I have the following transformer: 
class Clipper(BaseEstimator, TransformerMixin):
    def __init__(self,lower_limit,upper_limit):
        self.lower_limit = lower_limit
        self.upper_limit = upper_limit
        pass

    def clip(self,data):
        return np.clip(data,self.lower_limit,self.upper_limit)

    def fit(self, X, y=None):
        return self

    def transform(self,X,y=None):
        x = X.apply(lambda x: self.clip(x), axis=1)
        return x.values.reshape((-1, 1))

    def get_feature_names_out(self):
            pass
Questions:
1. Is there a way to write a converter which can accept the lower limit and upper limit arguments (similar to my init function)? 
2. In your example below, you use operator.inputs[0]. What other contents does the inputs list contain?

PR #1044, opened 2023-11-11
OrdinalEncoder conversion fixes

max-509:
Hello! Thank you for this awesome library that helps me to use sklearn models in highly loaded systems.

I think I found some problems for OrdinalEncoder conversion. Current implementation doesn't work with encoder that was fitted on data with different column datatypes. For example, if I pass to fit() function pandas dataframe with different column datatypes (np.int64, object), I get error when I try run ONNX model using onnxruntime framework because ONNX model contains inconsistent attributes for LabelEncoder operator.

My suggestion is removing cast to string for each input variable and pay attention not only to the `categories_` array dtype but also to the type of `categories_` values because in your `test_ordinal_encoder_mixed_string_int_drop()` function `categories_` array is object dtype, but values type is int.

In the end I want attach demo jupyter notebook and ONNX models with old and new conversions for problem demonstration.
[ordinal_encoder.zip](https://github.com/onnx/sklearn-onnx/files/13325554/ordinal_encoder.zip)

I will be waiting for comments on my pull request

### DIFF
--- a/skl2onnx/operator_converters/ordinal_encoder.py
+++ b/skl2onnx/operator_converters/ordinal_encoder.py
@@ -11,96 +11,81 @@ from ..proto import onnx_proto
 
 
 def convert_sklearn_ordinal_encoder(
-    scope: Scope, operator: Operator, container: ModelComponentContainer
+        scope: Scope, operator: Operator, container: ModelComponentContainer
 ):
     ordinal_op = operator.raw_operator
     result = []
-    concatenated_input_name = operator.inputs[0].full_name
-    concat_result_name = scope.get_unique_variable_name("concat_result")
+    input_idx = 0
+    dimension_idx = 0
+    for categories in ordinal_op.categories_:
+        if len(categories) == 0:
+            continue
 
-    if len(operator.inputs) > 1:
-        concatenated_input_name = scope.get_unique_variable_name("concatenated_input")
-        if all(
-            isinstance(inp.type, type(operator.inputs[0].type))
-            for inp in operator.inputs
-        ):
-            input_names = list(map(lambda x: x.full_name, operator.inputs))
+        current_input = operator.inputs[input_idx]
+        if current_input.get_second_dimension() == 1:
+            feature_column_name = current_input.onnx_name
+            input_idx += 1
         else:
-            input_names = []
-            for inp in operator.inputs:
-                if isinstance(inp.type, Int64TensorType):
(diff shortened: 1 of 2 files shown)


tests/test_sklearn_voting_classifier_converter.py
```diff
@@ -313,25 +324,10 @@ def test_voting_classifier_one_more_input(self):
                         estimators=[
                             (
                                 "est",
-                                Pipeline(
```
xadupre, line 316: Any particular reason why you changed this test?

max-509: Reason for this is strange 'SklearnConcat' conversion function that converts all inputs to first input type and concat inputs to single tensor. 

voting_classifier
This image is `test_voting_classifier_one_more_input()` with old pipeline. As you can see, SklearnConcat conversion cast all non string input nodes to StringTensorType and applies Concat operator. And after that LabelEncoder will fail because keys type is int64 by Original Encoder, but input variable type is string.

max-509: I think instead of using SklearnConcat converter you need add Concat/FeatureVectorizer operator to each converter that is needing single input in current implementation.
For example, SklearnVotingClassifier/Regressor doesn't need single input because each estimator in voting model is operator (Pipeline or Estimator) that can convert input variables to prediction/probability variables. 
And you can add Concat/FeatureVectorizer operator to, for example, SklearnRandomForest converter, SklearnLinearRegressor converter and etc.

xadupre: I see. But is the previous pipeline a valid pipeline from scikit-learn point of view? If so, we should do something about it as well.

max-509: I found solution for problem with this test case. Now OrdinalEncoder converter set keys types for LabelEncoder by input variable dtype, not by OrdinalEncoder `categories_` field type.
I have added these changes. In addition, I found some other bugs that I fixed in the last commit. They are described in the commit message

max-509:
@xadupre I have refactored the code via black

tests/test_sklearn_voting_classifier_converter.py
xadupre: Thanks for fixing it.

xadupre (approved):

PR #1132, opened 2024-10-18
`OrdinalEncoder` handle `encoded_missing_value` and `unknown_value`

LLukas22:
In the current version of `sklearn-onnx`, the `encoded_missing_value` and `unknown_value` parameters of the `OrdinalEncoder` in scikit-learn are not properly handled. Specifically, these parameters are ignored during the ONNX model conversion.

For example, if we create an `OrdinalEncoder` with `encoded_missing_value` set to `42` and fit it on the following data: `np.array([["a"], ["b"], ["c"], ["d"], [np.nan]], dtype=np.object_)`, scikit-learn produces the expected output: `[0, 1, 2, 3, 42]`. However, the converted ONNX model does not respect the `encoded_missing_value` parameter, leading to an unexpected result: `[0, 1, 2, 3, 4]`.

Similarly, the `unknown_value` parameter is also ignored during conversion, which affects the expected output. To address this issue, the `default_int64` parameter of the ONNX `LabelEncoder` needs to be set when an `unknown_value` is specified.

I have included some tests to demonstrate this behavior and implemented a simple fix to resolve the issue.

### DIFF
--- a/skl2onnx/operator_converters/ordinal_encoder.py
+++ b/skl2onnx/operator_converters/ordinal_encoder.py
@@ -106,8 +106,10 @@ def convert_sklearn_ordinal_encoder(
         if not np.isnan(ordinal_op.encoded_missing_value) and (
             isinstance(categories[-1], float) and np.isnan(categories[-1])
         ):
-            # sklearn always places np.nan as the last entry in its cathegories if it was in the training data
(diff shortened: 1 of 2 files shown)


xadupre (approved):

LLukas22:
Re-signed my commits, this should be good to go now 👍

xadupre (approved):

PR #1136, opened 2024-11-06
TargetEncoder support

boccaff:
This PR implements a converter and a shape calculator for the TargetEncoder class introduced in Scikit-learn 1.5. The code follows much of the implementation of the converter for Ordinal Encoder. 

A partial suit of tests is already implemented, but there is at least a couple of additional tests that I would like to add (missing values and using the smooth parameter from sklearn, even though I think it shouldn't matter).


boccaff:
I'll follow up with a cleaner and signed-off PR.

PR #1139, opened 2024-11-14
Create output_onnx_single_probability.py

ejosowitz:
Converts a RandomForestClassifier model to ONNX such that only a single (positive) probability is output


xadupre:
Let's keep it that way. I'll add extra comments to improve the rendering through sphinx-gallery.

xadupre (approved):

PR #1156, opened 2025-01-22
Test SelectKBest + StandardScaler pipeline

Pierre-Bartet:
When `SelectKBest` is used before a `StandardScaler` inside a pipeline with $k$ smaller than the number of features, the conversion fails:

``` python
model = make_pipeline(SelectKBest(k=3), StandardScaler(), LogisticRegression())
```

This PR fixes this problem by using the `concat` shape calculator for selectors and adding a correct one instead.


Pierre-Bartet:
@xadupre I fixed the issue (just for `SelectKBest`, not the other selectors), but before I go further I want to be sure I understand properly what was the initial intention of the code I changed.

All selectors such as `SelectKBest` use the `concat` shape calculator instead of something closer to the `array_feature_extractor` one. This look like a mistake to me, so I unregistered the selectors from the `concat` shape calculator, and use a slight variation of the `array_feature_extractor` one instead.

What bothers me is that in the end the `calculate_sklearn_array_feature_extractor` is never called during the test I added (and never was).

So my two questions are:

1. Was the use of the `concat` shape calculator for the selectors just a mistake, or was there some deeper intention here?
2. Is it intended that with or without my fix, `calculate_sklearn_array_feature_extractor` is not used when converting `SelectKBest` into an `array_feature_extractor`?

xadupre:
You should remove the register and run ``tests/test_sklearn_feature_selection_converters.py`` to see which unit tests require it. shape_calculator is a kind of legacy but it is used to compute the output shape and type of a model without converting it so that the conversion of the next estimator can start before the consersion of the previous is done. In practice, this never happens.

Pierre-Bartet:
No `tests/test_sklearn_feature_selection_converters.py` test fail when doing that, but other tests fail elsewhere.
So from what I understand, the answer to question 2. is "Yes".
I guess this is also the case for question 1, so I'll just proceed and do the same for all selectors.

Pierre-Bartet:
@xadupre done

PR #1159, opened 2025-01-23
Fix empty column selector

### LINKED ISSUE #1155: ColumnTransformer conversion is broken for common "corner cases"
`ColumnTransformer` conversion raises an error [here](https://github.com/onnx/sklearn-onnx/blob/ffb17c505829cde7e4edbc3f123a0730db837727/skl2onnx/_parse.py#L261) when fitted on data for which one of the column selector has no matching column.  
This is due to the fact that in that case `inputs` will be an empty list, so that the `guess_tensor_type(inputs[0].type)` line fails.

For example, the following `ColumnTransformer`, taken from the `sklearn` documentation, can't be converted to `ONXX` when the input dataframe happens to contain no numerical column or no string column:

``` python
ct = make_column_transformer(
    (StandardScaler(), make_column_selector(dtype_include=np.number)),  # rating
    (OneHotEncoder(), make_column_selector(dtype_include=object)),  # city
)

X = pd.DataFrame({
    'city': ['London', 'London', 'Paris', 'Sallisaw'],
    # 'rating': [5, 3, 4, 5], 
}) 

ct.fit(X)

to_onnx(ct, X)
```

This can be solved by checking for empty inputs inside or before calling `_parse_sklearn_column_transformer`.

Pierre-Bartet:
Fixes #1155 by skipping transforms with no corresponding columns at all, and add the corresponding test.


Pierre-Bartet:
@xadupre this one is ready for review (and I turned the non ready one into a draft).

Pierre-Bartet:
Are the workflows run automatically when pushing a new commit or do you have to approve each run each time ?

xadupre:
I have to approve.

Pierre-Bartet:
Sorry for not running Black before pushing then ^^

xadupre (approved):

xadupre:
Should I merge?

Pierre-Bartet:
This is a one liner fix and all the existing and added tests pass so yes!

PR #1165, opened 2025-02-04
Fix KNNImputer

xadupre:


ashtarimo:
KNNImputer converter doesn't generate the same values for the imputed missing values. I am using:

onnxruntim 1.18.1, onnx  1.17.0, and skl2onnx 1.17.0

Here is the code:


import numpy as np
import onnx
import skl2onnx
from sklearn.impute import KNNImputer

np.random.seed(42)
data = np.random.randn(100, 10)
data[99, 9] = np.nan
imputer = KNNImputer(n_neighbors= 5)
imputer.fit(data)
dataft = imputer.transform(data)
initial_type = [('float_input', FloatTensorType([None, data.shape[1]]))]
onnx_model = convert_sklearn(imputer, initial_types=initial_type)
with open("knn_imputer.onnx", "wb") as f:
    f.write(onnx_model.SerializeToString())
sess = rt.InferenceSession("knn_imputer.onnx")
input_data = data.astype(np.float32)
input_name = sess.get_inputs()[0].name
output_name = sess.get_outputs()[0].name
res = sess.run([output_name], {input_name: input_data})

The KNNIMputer generates value of -0.67440104 for the missing value, and the onnx model, generates -1.0071397 for the same missing value

PR #1179, opened 2025-05-05
Use onnx-weekly in CI, remove dependency on onnxconveter-common

xadupre:


khoover:
What was motivation for this? This breaks using the onnxmltools parsers for LightGBM, XGBoost, etc. within a scikit Pipeline. The converters and calculators work fine for now thanks to the duck typing, but without that common dependency I suspect they'll break too eventually.

xadupre:
> What was motivation for this? This breaks using the onnxmltools parsers for LightGBM, XGBoost, etc. within a scikit Pipeline. The converters and calculators work fine for now thanks to the duck typing, but without that common dependency I suspect they'll break too eventually.

I'll be fixing that soon. I removed the dependency because onnxconverter-common was used only for those common types and it was not really needed beside that. There are only two dependencies now, onnx and scikit-learn. It seemed easier to maintain on the long run.

PR #1181, opened 2025-05-14
Remove import of `split_complex_to_pairs`

justinchuby:
It seems to be unused. The function is made internal in onnx 1.18.


xadupre:
Replaced by #1182

PR #1186, opened 2025-05-26
Migrate to use pyproject to simplify setup

justinchuby:
Removed logic in setup.py that parses version etc. and instead use a more simplified pyproject.toml config.

@xadupre


pyproject.toml
```diff
@@ -1,3 +1,44 @@
...
+    "Intended Audience :: Developers",
+    "Operating System :: MacOS :: MacOS X",
+    "Operating System :: Microsoft :: Windows",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
```
justinchuby, line 27: Added new versions

PR #1205, opened 2025-08-01
On inputs/outputs validation errors, allow caller to inspect them

amol-:
At the moment, when parsing a SciKit-Learn pipeline, the inputs validation throws a generic `RuntimeError`.

This prevents the caller to understand what was wrong, especially in the case where the Pipeline involved a Dataframe and thus multiple separate columns existed but the ONNX model wants a single concatenated input tensor.

In [posit-dev/orbital ](https://github.com/posit-dev/orbital) we were forced to inspect the error message text to understand what was the actual error, which is not a robust solution.

```
        # Test if the operator can handle the individual inputs we provided
        try:
            first_operator.infer_types()
            # If infer_types() succeeds, the operator accepts the inputs the user provided
            return False
        except RuntimeError as err:
            if "at most 1 input" in str(err):
                # If infer_types() fails with "at most 1 input", the operator needs concatenated inputs
                # This is the best we can do as SKL2ONNX doesn't tell us how many inputs it expects.
                # And the `check_input_and_output_numbers` function always throws a RuntimeError
                return True
            return False
```

This PR does:

1. Create separate exception classes for input errors, output errors and type errors. So that the caller can distinguish between the 3 possible kind of errors.
2. Provide this exception classes with the data necessary to understand what was the expectation from the caller.
3. Fully retain backward compatibility by inheriting the exception from `RuntimeError` and keeping the current order of error messages (input first, output only if input check passes).


xadupre:
Can you add a line in CHANGELOGS.md?

amol-:
@xadupre I think this should be good for a second pass as soon as you have time to take a look again 🙂

amol-:
> To fix the style:
> 
> ```
> black .
> ruff check .
> ```

Oh thanks, I did run `ruff format` but I guess black might have been configured slightly differently

xadupre:
> > To fix the style:
> > ```
> > black .
> > ruff check .
> > ```
> 
> Oh thanks, I did run `ruff format` but I guess black might have been configured slightly differently

It should be the default settings. I don't like spending time on these settings.

xadupre (approved):

PR #1249, opened 2026-05-15
Fix Pipeline subclass conversion by walking MRO in _get_allowed_options

### LINKED ISSUE #1248: Pipeline subclasses raise `NotImplementedError` during conversion since 1.17.0 (regression from PR #1072)
## Summary

Converting a `Pipeline` subclass that contains a classifier raises `NotImplementedError`
in skl2onnx ≥ 1.17.0. The same code works correctly in ≤ 1.16.0.

## Reproduction

▶️ [Run in Google Colab](https://colab.research.google.com/gist/slam2sense/18db6bc46dc9ab01ccc6d638b79d2bf6/pipeline_subclass_repro.ipynb)

```python
import numpy as np
from sklearn.pipeline import Pipeline
from sklearn.preprocessing import StandardScaler
from sklearn.linear_model import LogisticRegression

class MyPipeline(Pipeline):
    pass

pipe = MyPipeline([("scaler", StandardScaler()), ("clf", LogisticRegression())])
pipe.fit(np.random.randn(50, 4), np.random.randint(0, 2, 50))

convert_sklearn(pipe, initial_types=[("X", FloatTensorType([None, 4]))])
```

**Error (≥ 1.17.0):**
```
NotImplementedError: No registered models, no known allowed options for model 'MyPipeline'.
```

**Works in ≤ 1.16.0.**

## Root cause

PR #1072 ("Investigate issue 1069", merged April 2024) added `zipmap` option propagation
inside `_parse_sklearn_pipeline`
([`_parse.py:305–310`](https://github.com/onnx/sklearn-onnx/blob/1.20.0/skl2onnx/_parse.py#L305-L310)):

```python
for step in model.steps:
    if (
        scope.options is None
        or (id(step[1]) not in scope.options and type(step[1]) not in scope.options)
    ) and is_classifier(step[1]):
        options = scope.get_options(model, {"zipmap": True})  # model = the Pipeline
        scope.add_options(id(step[1]), options)
    inputs = _parse_sklearn(scope, step[1], inputs, custom_parsers=custom_parsers)
```

Note that `_parse_sklearn` already handles subclasses correctly: at
[`_parse.py:709`](https://github.com/onnx/sklearn-onnx/blob/1.20.0/skl2onnx/_parse.py#L709)
it uses `isinstance` to dispatch any `Pipeline` subclass to `_parse_sklearn_pipeline`:

```python
elif isinstance(model, pipeline.Pipeline):
    parser = sklearn_parsers_map[pipeline.Pipeline]
    outputs = parser(scope, model, inputs, custom_parsers=custom_parsers)
```

So the dispatch layer is fine. The problem is that once inside `_parse_sklearn_pipeline`,
the new zipmap path calls `scope.get_options(model, {"zipmap": True})`, which in turn
calls `_get_allowed_options(model)` in `_topology.py`
([L892](https://github.com/onnx/sklearn-onnx/blob/1.20.0/skl2onnx/common/_topology.py#L892)):

```python
def _get_allowed_options(self, model, fail=True):
    if self.registered_models is not None:
        if type(model) not in self.registered_models["aliases"]:  # strict type() lookup
            if fail:
                raise NotImplementedError(
                    "No registered models, no known allowed options "
                    "for model '{}'.".format(model.__class__.__name__)
                )
            return {}
        alias = self.registered_models["aliases"][type(model)]
        ...
```

`registered_models["aliases"]` is `sklearn_operator_name_map` (populated in
`_supported_operators.py`), which maps **exact concrete classes** — only
`sklearn.pipeline.Pipeline` itself is registered, not subclasses.

**Before PR #1072**, `_parse_sklearn_pipeline` was just:

```python
for step in model.steps:
    inputs = _parse_sklearn(scope, step[1], inputs, custom_parsers=custom_parsers)
return inputs
```

`_get_allowed_options` was never called with the Pipeline object itself, so subclasses
worked. The new zipmap propagation path introduced in #1072 triggers it and raises for
any `Pipeline` subclass.

## Workaround

▶️ [Run in Google Colab](https://colab.research.google.com/gist/slam2sense/78ffb07d894ef8f7fc89dce636fcbab9/pipeline_subclass_workaround.ipynb)

Until the fix lands, manually register the subclass in `sklearn_operator_name_map` inside `__init__`:

```python
from skl2onnx._supported_operators import sklearn_operator_name_map

class MyPipeline(Pipeline):
    def __init__(self, steps, **kwargs):
        super().__init__(steps, **kwargs)
        sklearn_operator_name_map[type(self)] = "SklearnPipeline"
```

## Suggested fix

Two options, in increasing order of robustness:

**Option A** — pass `fail=False` at the call site so an unregistered subclass silently
returns `{}` and falls back to the default (`zipmap=True`), rather than raising:

```python
# _parse.py:309
options = scope.get_options(model, {"zipmap": True}, fail=False)
```

**Option B** — make `_get_allowed_options` walk the MRO so subclasses inherit their
parent's registration transparently:

```python
# _topology.py:890
def _get_allowed_options(self, model, fail=True):
    if self.registered_models is not None:
        for cls in type(model).__mro__:
            if cls in self.registered_models["aliases"]:
                alias = self.registered_models["aliases"][cls]
                conv = self.registered_models["conv"][alias]
                return conv.get_allowed_options()
        if fail:
            raise NotImplementedError(
                "No registered models, no known allowed options "
                "for model '{}'.".format(model.__class__.__name__)
            )
        return {}
    raise NotImplementedError(
        "No registered models, no known allowed options "
        "for model '{}'.".format(model.__class__.__name__)
    )
```

Option B is more complete — it would fix not just this path but any other place in the
codebase where a `Pipeline` subclass (or any other subclass of a registered model) hits
the strict `type()` lookup. Option A is a targeted one-liner scoped to the regression.

## Version info

- Broken: skl2onnx ≥ 1.17.0 (regression introduced by PR #1072; current: 1.21.0)
- Working: skl2onnx ≤ 1.16.0

slam2sense:
## Summary

- Fixes #1248: `Pipeline` subclasses raised `NotImplementedError` during conversion since 1.17.0 (regression introduced by PR #1072)
- Replaces the strict `type(model)` lookup in `_get_allowed_options` with an MRO walk, so subclasses inherit their parent's registered options transparently
- Fix applies to `Pipeline` and any other registered model subclass that hits this path

## Root cause

PR #1072 added zipmap option propagation inside `_parse_sklearn_pipeline`, calling `scope.get_options(model, {"zipmap": True})` with the pipeline itself. `_get_allowed_options` used a strict `type(model)` check against `sklearn_operator_name_map`, which only registers exact concrete classes — not subclasses.

## Changes

- `skl2onnx/common/_topology.py`: replace `type(model) not in aliases` guard with a `for cls in type(model).__mro__` walk
- `tests/test_issues_2026.py`: regression test for `Pipeline` subclass conversion

## Test

```python
class MyPipeline(Pipeline):
    pass

pipe = MyPipeline([("scaler", StandardScaler()), ("clf", LogisticRegression())])
pipe.fit(X, y)
convert_sklearn(pipe, initial_types=[("X", FloatTensorType([None, 4]))])  # previously raised NotImplementedError
```

### DIFF
--- a/skl2onnx/common/_topology.py
+++ b/skl2onnx/common/_topology.py
@@ -7,6 +7,7 @@ import pprint
 from logging import getLogger
 from collections import OrderedDict
 import numpy as np
+from sklearn import pipeline
 from onnx import onnx_pb as onnx_proto
 from onnx.helper import (
     make_graph,
@@ -889,21 +890,20 @@ class Scope:
 
     def _get_allowed_options(self, model, fail=True):
         if self.registered_models is not None:
+            aliases = self.registered_models["aliases"]
             model_cls = type(model)
-            for cls in model_cls.__mro__:
-                if cls in self.registered_models["aliases"]:
-                    if cls is not model_cls:
-                        warnings.warn(
-                            "Class '{}' is not registered; using converter "
-                            "registered for parent class '{}'.".format(
-                                model_cls.__name__, cls.__name__
-                            ),
-                            UserWarning,
-                            stacklevel=0,
-                        )
-                    alias = self.registered_models["aliases"][cls]
-                    conv = self.registered_models["conv"][alias]
-                    return conv.get_allowed_options()
+            lookup_cls = None
+            if model_cls in aliases:
+                lookup_cls = model_cls
+            elif (
+                isinstance(model, pipeline.Pipeline)
+                and pipeline.Pipeline in aliases
+            ):
+                lookup_cls = pipeline.Pipeline
+            if lookup_cls is not None:
+                alias = aliases[lookup_cls]
+                conv = self.registered_models["conv"][alias]
+                return conv.get_allowed_options()
             if fail:
                 raise NotImplementedError(
                     "No registered models, no known allowed options "
--- a/skl2onnx/common/_topology.py
+++ b/skl2onnx/common/_topology.py
@@ -889,8 +889,18 @@ class Scope:
 
     def _get_allowed_options(self, model, fail=True):
         if self.registered_models is not None:
-            for cls in type(model).__mro__:
+            model_cls = type(model)
+            for cls in model_cls.__mro__:
                 if cls in self.registered_models["aliases"]:
+                    if cls is not model_cls:
+                        warnings.warn(
+                            "Class '{}' is not registered; using converter "
+                            "registered for parent class '{}'.".format(
+                                model_cls.__name__, cls.__name__
+                            ),
+                            UserWarning,
+                            stacklevel=0,
+                        )
                     alias = self.registered_models["aliases"][cls]
                     conv = self.registered_models["conv"][alias]
                     return conv.get_allowed_options()
--- a/skl2onnx/common/_topology.py
+++ b/skl2onnx/common/_topology.py
@@ -889,17 +889,17 @@ class Scope:
 
     def _get_allowed_options(self, model, fail=True):
         if self.registered_models is not None:
-            if type(model) not in self.registered_models["aliases"]:
-                if fail:
-                    raise NotImplementedError(
-                        "No registered models, no known allowed options "
-                        "for model '{}'.".format(model.__class__.__name__)
-                    )
-                return {}
-            alias = self.registered_models["aliases"][type(model)]
-            conv = self.registered_models["conv"][alias]
-            allowed = conv.get_allowed_options()
-            return allowed
+            for cls in type(model).__mro__:
+                if cls in self.registered_models["aliases"]:
+                    alias = self.registered_models["aliases"][cls]
+                    conv = self.registered_models["conv"][alias]
+                    return conv.get_allowed_options()
+            if fail:
+                raise NotImplementedError(
+                    "No registered models, no known allowed options "
+                    "for model '{}'.".format(model.__class__.__name__)
+                )
+            return {}
         raise NotImplementedError(
             "No registered models, no known allowed options "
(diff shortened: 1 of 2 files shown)


skl2onnx/common/_topology.py
```diff
@@ -889,17 +889,17 @@ def declare_local_operator(self, type, raw_model=None):
...
-                    raise NotImplementedError(
-                        "No registered models, no known allowed options "
-                        "for model '{}'.".format(model.__class__.__name__)
-                    )
-                return {}
-            alias = self.registered_models["aliases"][type(model)]
-            conv = self.registered_models["conv"][alias]
-            allowed = conv.get_allowed_options()
-            return allowed
+            for cls in type(model).__mro__:
```
xadupre: I suggest adding a warning to tell the user the converted class is one of the parent, not the class itself.

slam2sense: one detail I just noticed, when the model is not registered, if it is a `Pipeline`, [the parsing fallback](https://github.com/onnx/sklearn-onnx/blob/main/skl2onnx/_parse.py#L700-L716) is to use the `pipeline.Pipeline` parser and not traverse the mro.


Should we rather do that to be consistent?

xadupre: I prefer. Using __mro__ is quite unusual.

xadupre (approved):

PR #1254, opened 2026-07-16
Fix tree converters to propagate learned NaN routing (missing_go_to_left)

### LINKED ISSUE #1253: Bug: Decision tree / forest converters drop scikit-learn's learned NaN routing (missing_go_to_left)
Tree-based estimators route NaN per-node via `missing_go_to_left` (learned at fit). `add_tree_to_attribute_pairs` never reads it, so `nodes_missing_value_tracks_true` is emitted all-zeros and every NaN takes the false branch in ONNX Runtime. Predictions diverge from scikit-learn on rows with missing values; clean rows match. `add_tree_to_attribute_pairs_hist_gradient_boosting` in the same file already copies the attribute, so `HistGradientBoosting*` is unaffected, which is the precedent for the fix.

Reproduction (pin onnxruntime<=1.25.1 because 1.26.0 seems to have a binary-complement regression that I don't quite understand yet):

```python
import numpy as np
import onnxruntime as rt
from sklearn.ensemble import RandomForestClassifier

rng = np.random.default_rng(0)
n, d = 600, 6
X = rng.normal(size=(n, d)).astype(np.float32)
y = (X[:, 0] + 0.5 * X[:, 1] > 0).astype(int)
X[rng.random((n, d)) < 0.15] = np.nan  # 15% missing per column

clf = RandomForestClassifier(n_estimators=50, max_depth=5, random_state=0).fit(X, y)
assert any(np.asarray(e.tree_.missing_go_to_left).any() for e in clf.estimators_)

onx = convert_sklearn(clf, initial_types=[("input", FloatTensorType([None, d]))], options={"zipmap": False})
sess = rt.InferenceSession(onx.SerializeToString(), providers=["CPUExecutionProvider"])
onnx_proba = sess.run(None, {"input": X})[1]
skl_proba = clf.predict_proba(X)
diff = np.abs(onnx_proba - skl_proba).max(axis=1)
nan_rows = np.isnan(X).any(axis=1)
print(f"max diff on NaN rows:   {diff[nan_rows].max():.4f}")   # observed ~0.70
print(f"max diff on clean rows: {diff[~nan_rows].max():.4f}")  # observed 0.0000
```

Versions: skl2onnx 1.21.0 (and main), scikit-learn 1.9.0, onnxruntime 1.25.1, onnx 1.21.0. Also reproduces with onnx 1.22.0.

A ready-to-run version of this reproduction, including a script that builds the exact venv with uv in one command, is in my fork branch https://github.com/warmlogic/sklearn-onnx/tree/matt/tree-nan-repro/reproductions (`reproductions/` on the `matt/tree-nan-repro` branch). It uses a single venv (numpy, scikit-learn, and onnx at their latest release, onnxruntime pinned).

Expected: ONNX predictions match scikit-learn on NaN rows, as they already do for HistGradientBoosting*.

warmlogic:
## What this fixes

Decision tree, random forest, and extra trees converters drop scikit-learn's learned NaN routing. `add_tree_to_attribute_pairs` in `skl2onnx/common/tree_ensemble.py` never reads `tree_.missing_go_to_left`, so the emitted ONNX node attribute `nodes_missing_value_tracks_true` is always zero. In ONNX Runtime, every NaN input then takes the false branch regardless of what scikit-learn actually learned. Predictions match exactly on clean rows and diverge sharply on rows with a missing value.

`add_tree_to_attribute_pairs_hist_gradient_boosting`, in the same file, already reads the equivalent attribute correctly, so `HistGradientBoosting*` converters are unaffected. This PR brings the plain-tree path in line with that existing precedent.

Fixes #1253

## Before and after

Using a `RandomForestClassifier` and a `DecisionTreeClassifier` fit on data with about 15% NaNs injected (full repro at [`reproductions/`](https://github.com/warmlogic/sklearn-onnx/tree/matt/tree-nan-repro/reproductions) on the `matt/tree-nan-repro` branch of this fork):

|            | max diff, NaN rows        | max diff, clean rows |
| ---------- | ------------------------- | -------------------- |
| Before fix | RF 0.699077 / DT 0.951220 | 0.000000             |
| After fix  | 0.000000                  | 0.000000             |

## Changes

- `skl2onnx/common/tree_ensemble.py`: `add_tree_to_attribute_pairs` now reads `tree.missing_go_to_left` per node and passes it through as `nodes_missing_value_tracks_true`, mirroring the existing HGB function. Trees fitted by scikit-learn versions without the attribute fall back to today's behavior (0), so nothing regresses for older scikit-learn.
- `tests/test_sklearn_random_forest_converters.py`: added NaN round-trip tests for `RandomForestClassifier`, `RandomForestRegressor`, and `ExtraTreesClassifier`.
- `tests/test_sklearn_decision_tree_converters.py`: added NaN round-trip tests for `DecisionTreeClassifier` and `DecisionTreeRegressor`.

All five new tests are gated on a runtime `hasattr(tree_, "missing_go_to_left")` check rather than a hardcoded scikit-learn version, matching how the fix itself detects support.

## Test plan

- [x] `pytest tests/test_sklearn_decision_tree_converters.py tests/test_sklearn_random_forest_converters.py -k nan`: 5 new tests pass
- [x] Confirmed the 5 new tests fail against the pre-fix source (reverted only `tree_ensemble.py`, kept the new tests) and pass again once restored, so they are real regression tests
- [x] `ruff check .` and `black --check .` clean on changed files
- [x] Full `pytest tests` against latest numpy/scikit-learn/onnx/onnxruntime: no new failures introduced. 18 pre-existing failures unrelated to this change (verified against unmodified main), all caused by this project's converters not yet supporting scikit-learn 1.9.0 internals (HistGradientBoosting node layout, SVM `probA_`/`probB_` deprecation). Out of scope for this PR.


xadupre (approved):

xadupre:
I fixed CI but now the tests should be excluded from the job testing versions of scikit-learn not supporting nans.

warmlogic:
> I fixed CI but now the tests should be excluded from the job testing versions of scikit-learn not supporting nans.

@xadupre Thanks for fixing CI, you're right about the exclusions. My version check looked at the wrong thing.

scikit-learn 1.3 added `tree_.missing_go_to_left`, so the `hasattr` check passed on every job from that version. Whether `fit` accepts NaN actually depends on the splitter: through 1.5.x, [`sklearn/tree/_classes.py`](https://github.com/scikit-learn/scikit-learn/blob/1.5.2/sklearn/tree/_classes.py#L1085) sets `allow_nan = self.splitter == "best" and ...`, and [1.6.0](https://github.com/scikit-learn/scikit-learn/blob/1.6.0/sklearn/tree/_classes.py#L1727) extended it to `"random"`. `ExtraTreesClassifier` defaults to `splitter="random"`, so `test_model_extra_trees_classifier_nan` got past my incorrect check and then failed inside `fit` with `ValueError: Input X contains NaN` on scikit-learn 1.5.2.

I replaced it with `_sklearn_version()` comparisons (the same approach the `HistGradientBoosting` NaN tests use): `>= 1.3` for the DT and RF tests, `>= 1.6` for ET. Tests with the pinned versions from three of the jobs now pass/skip as expected.

warmlogic:
The failing ruff check here is not from this PR. ruff 0.16.0 added a new rule that fires on `docs/tutorial/plot_transformer_discrepancy.py`, so `main` and every open PR are red for the same reason. I put the fix in d350bf0c to unblock this one, and PR #1262 fixes it properly on `main`.

`CI Win/Mac` is red here too, but it has been failing on `main` since March for an unrelated reason.

Filed issue #1261 with both, plus a few other CI problems I ran into along the way.

